PR #10, opened 2021-01-22
Add error handling to main()

### LINKED ISSUE #2: Better error handling
Right now the `main` function is full of `.unwrap()` calls. Every time something minor happens (like the user tries to pull a WASM module from a wrong HTTP endpoint), the policy-server panics and prints a message.

We should do better error handling:
  * Do not cause the main program to panic
  * Catch the error, print it to STDERR and then exit with 1

Tip: take a look at the [rust book for cli apps](https://rust-cli.github.io/book/tutorial/errors.html#no-need-to-panic) for inspiration.

cmurphy:
Where user input could cause an error, replace calls to unwrap() with
error messages and an exit code rather than panicking.

Fixes #2

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,6 +2,7 @@ use clap::{App, Arg};
 use hyper::service::{make_service_fn, service_fn};
 use hyper::Server;
 use std::{net::SocketAddr, thread};
+use std::process;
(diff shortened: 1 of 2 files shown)


flavio:
LGTM, there are still a couple of `unwrap` calls when looking at the values provided by `clap`, but I think they are reasonable.

PR #11, opened 2021-01-25
Implement fetching from an artifact-enabled OCI registry

ereslibre:
By using a `registry://` schema on the `--wasm-uri` argument, it's
possible to fetch the WASM binary from an artifact enabled OCI
registry.

Fixes https://github.com/chimera-kube/policy-server/issues/1

### DIFF
--- a/src/wasm_fetcher/mod.rs
+++ b/src/wasm_fetcher/mod.rs
@@ -1,14 +1,16 @@
 use anyhow::{anyhow, Result};
-use hyper::Uri;
+use url::Url;
 use std::boxed::Box;
 
 pub mod fetcher;
 mod https;
 mod local;
+mod registry;
 
 use crate::wasm_fetcher::fetcher::Fetcher;
 use crate::wasm_fetcher::https::Https;
 use crate::wasm_fetcher::local::Local;
+use crate::wasm_fetcher::registry::Registry;
 
 // Helper function, takes the URL of the WASM module and allocates
 // the right struct to interact with it
@@ -19,7 +21,7 @@ pub(crate) fn parse_wasm_url(
 ) -> Result<Box<dyn Fetcher>> {
   // we have to use url::Url instead of hyper::Uri because the latter one can't
   // parse urls like file://
-  let parsed_url = match url::Url::parse(url) {
+  let parsed_url: Url = match url::Url::parse(url) {
     Ok(u) => u,
     Err(e) => {
       return Err(anyhow!("Invalid WASI url: {}", e));
@@ -28,8 +30,8 @@ pub(crate) fn parse_wasm_url(
 
   match parsed_url.scheme() {
     "file" => Ok(Box::new(Local::new(parsed_url.path())?)),
-    "http" => Ok(Box::new(Https::new(url.parse::<Uri>()?, remote_insecure)?)),
-    "https" => Ok(Box::new(Https::new(url.parse::<Uri>()?, remote_insecure)?)),
+    "http" | "https" => Ok(Box::new(Https::new(url.parse::<Url>()?, remote_insecure)?)),
+    "registry" => Ok(Box::new(Registry::new(parsed_url, remote_non_tls)?)),
     _ => Err(anyhow!("unknown scheme: {}", parsed_url.scheme())),
   }
 }
--- a/src/wasm_fetcher/registry.rs
+++ b/src/wasm_fetcher/registry.rs
@@ -1,0 +1,66 @@
+use anyhow::{anyhow, Result};
+use async_std::fs::File;
+use async_std::prelude::*;
+use async_trait::async_trait;
+use oci_distribution::Reference;
+use oci_distribution::client::{Client, ClientConfig, ClientProtocol};
+use oci_distribution::secrets::RegistryAuth;
+use std::str::FromStr;
+use tokio_compat_02::FutureExt;
+use url::Url;
+
+use crate::wasm_fetcher::fetcher::Fetcher;
+
+// Struct used to reference a WASM module that is hosted on an OCI registry
+pub(crate) struct Registry {
+  // full path to the WASM module
+  destination: String,
+  // url of the remote WASM module
+  wasm_url: String,
+  // whether TLS should be skipped
+  skip_tls: bool,
+}
+
+impl Registry {
+    pub(crate) fn new(url: Url, skip_tls: bool) -> Result<Registry> {
+        match url.path().rsplit('/').next() {
+            Some(image_ref) => {
+                let wasm_url = url.to_string();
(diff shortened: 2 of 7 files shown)


src/wasm_fetcher/mod.rs
```diff
@@ -28,8 +30,8 @@ pub(crate) fn parse_wasm_url(
 
   match parsed_url.scheme() {
     "file" => Ok(Box::new(Local::new(parsed_url.path())?)),
-    "http" => Ok(Box::new(Https::new(url.parse::<Uri>()?, remote_insecure)?)),
-    "https" => Ok(Box::new(Https::new(url.parse::<Uri>()?, remote_insecure)?)),
+    "http" | "https" => Ok(Box::new(Https::new(url.parse::<Url>()?, remote_insecure)?)),
+    "registry" => Ok(Box::new(Registry::new(parsed_url)?)),
```
flavio, line 34: The registry handler must take these flags at creation time:
  * `remote_insecure`: the remote OCI registry is secured with a self-signed tls key
  * `remote_non_tls`: the remote OCI registry is not secured by tls, it's served over plain http

src/wasm_fetcher/registry.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+#[async_trait]
+impl Fetcher for Registry {
+    async fn fetch(&self) -> Result<String> {
+        let mut client = Client::new(ClientConfig{
+            protocol: ClientProtocol::Https,
+        });
+        let reference = Reference::from_str(self.wasm_url.as_str()).unwrap();
+        let image_content = client.pull_image(&reference, &RegistryAuth::Anonymous).compat().await?.content;
```
flavio, line 49: Can you add a comment explaining why we are using the `compact()` function? Just to leave a note about the discrepancies between tokio versions and what we did to sort them out.

src/wasm_fetcher/registry.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+#[async_trait]
+impl Fetcher for Registry {
+    async fn fetch(&self) -> Result<String> {
+        let mut client = Client::new(ClientConfig{
+            protocol: ClientProtocol::Https,
+        });
+        let reference = Reference::from_str(self.wasm_url.as_str()).unwrap();
```
flavio, line 48: Instead of `unwrap` we should just bubble up the error via `?`

flavio (changes requested):
There are minor things to improve/implement.

I've also a question about the library we are using to pull from the OCI registry. How is authentication handled? Does it read the tokens from `~/.docker/config.json`?

src/wasm_fetcher/mod.rs
ereslibre: I don't see a straightforward way to configure this so that `oci-distribution` honors these settings. On their `ClientConfig` struct there's only the `protocol` that can be configured, but that only signals whether you want to use Http, Https, or Https except for a fixed number of hosts that will use Http.

Besides this, we probably want to check `remote_non_tls` again because we are setting it, but no backend is honoring it (the HTTP one is not honoring it either).

ereslibre: PR updated so it honors the `remote_non_tls` setting. I don't find a way to honor `remote_insecure` though (HTTPS but ignoring the presented certificate).

As a second question, given that the user can provide `--wasm-uri` with `http`, `https` and `registry` I specially wonder about how `http` and `https` interact from the UX perspective with `wasm-remote-non-tls` specially and `wasm-remote-insecure` as well, given that they are somehow tied to what schema the user provides on `--wasm-uri`.

ereslibre:
> How is authentication handled? Does it read the tokens from `~/.docker/config.json`?

No, there is no machinery like that implemented in this crate (at least yet, not sure if it's a goal for them). On certain calls you can provide authentication information, either by accessing as anonymous, or as a basic auth user and password combination.

src/wasm_fetcher/mod.rs
flavio: > Besides this, we probably want to check remote_non_tls again because we are setting it, but no backend is honoring it (the HTTP one is not honoring it either).

The `https` fetcher handles the `remote_non_tls` parameter automatically. If the URL scheme is `http://`, it will pull from a non-tls terminated server.

flavio: > PR updated so it honors the remote_non_tls setting. I don't find a way to honor remote_insecure though (HTTPS but ignoring the presented certificate).

I fear you will have to look into the OCI library and see which http client it uses. Based on that you will have to do some dancing around to make it ignore the remote tls certificate when this is not valid. I suspect this will end up with an upstream contribution

flavio: > As a second question, given that the user can provide --wasm-uri with file, http, https and registry I specially wonder about how http and https interact from the UX perspective with wasm-remote-non-tls specially and wasm-remote-insecure as well, given that they are somehow tied to what schema the user provides on --wasm-uri.

Let's park that for a short while, I think we will have to drop these flags once we move to support multiple WASM policies per server (see https://github.com/chimera-kube/policy-server/issues/5).

src/wasm_fetcher/registry.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[async_trait]
+impl Fetcher for Registry {
+    async fn fetch(&self) -> Result<String> {
+        let mut client = Client::new(ClientConfig{
+            protocol: if self.skip_tls { ClientProtocol::Http } else { ClientProtocol::Https },
+        });
+        let reference = Reference::from_str(self.wasm_url.as_str())?;
+        let image_content = client
+            .pull_image(&reference, &RegistryAuth::Anonymous)
```
flavio, line 53: How can we pull for a registry with authorization turned on?

flavio (changes requested):
Everything looks good, just the question about pull from authorization only servers

flavio:
> > How is authentication handled? Does it read the tokens from `~/.docker/config.json`?
> 
> No, there is no machinery like that implemented in this crate (at least yet, not sure if it's a goal for them). On certain calls you can provide authentication information, either by accessing as anonymous, or as a basic auth user and password combination.

Ouch, I think we can merge this PR and then create immediately an issue to ensure we can pull from auth protected registries.

ereslibre:
> Ouch, I think we can merge this PR and then create immediately an issue to ensure we can pull from auth protected registries.

+1, I am taking care of creating the issue.

src/wasm_fetcher/registry.rs
ereslibre: I created https://github.com/chimera-kube/policy-server/issues/12 with a very general intent and with some questions that need answer when implementing.

PR #15, opened 2021-01-27
Add TLS support for server acceptor

### LINKED ISSUE #4: Server: perform tls termination
Right now the server started by policy-server listens only over HTTP. We should make it easy to start a HTTPS server.

The cli app has the flags required to provide the key and cert to be used by the http server, but there's no code handling that.

We are using hyper to implement our web server, the tls termination must be done using [hyper-tls](https://crates.io/crates/hyper-tls). This is already used by the WASM HTTPS fetcher.

Some extra information: the WASM HTTPS fetcher is using hyper-tls instead of hyer-rustls because the latter one is not flexible enough for its use case. The hyper-rustls crate doesn't allow to create a HttpsConnector that accepts certificates signed by unknown CAs, plus it doesn't have a way to add custom CAs to verify them.
Now, for the https server these features are not relevant, but we don't want to have two tls crates as dependency.

cmurphy:
Use the tokio-native-tls crate to implement the Accept trait for
hyper::server so that it can accept incoming HTTPS requests.

Changes the --cert-file/--cert-key flags to
--client-certificate/--client-key to be consistent with the
kubeconfig naming. If neither are provided, a regular HTTP server is
created.

Fixes #4


flavio (changes requested):
This is one of the times when I feel so angry at Rust... I mean, with Go we would have done all of that in 10 minutes.

Aside from that:
  * Can you please change the behavior of the code to start a HTTPS server when the key/identity is provided; otherwise just start a HTTP one. Having a HTTP server is handy when doing load testing and similar.
  * Can we go ahead and get it merged with support for PEM files instead of pkcs12? After some googling I found https://github.com/sfackler/rust-native-tls/issues/27#issuecomment-289324928 Maybe this just works...

cmurphy:
`SecIdentityCreateWithCertificate` seems to be specific to Apple, but I was able to use the `openssl` crate to convert X.509 keys to PKCS#12. I added back the ability to run with HTTP only. The `make_svc` closure gets a little weird with the different `Server` types and I had to define it twice to get it to compile, open to suggestions on how to avoid that repetition.

cmurphy:
Changed this up to use a macro to DRY up the service closure without the compiler complaining about bound variables...

ereslibre:
Thank you @cmurphy, one question.

> Changes the --cert-file/--cert-key flags to
--client-certificate/--client-key to be consistent with the
kubeconfig naming.

Being the kubeconfig used by clients I think the naming makes sense there so they can be provided with client certs, but in the server side I think the naming of the argument is misleading, since this certificate and key are used for serving and accepting connections, and as such they should be server side certificates ensuring the "identity" of the sever, not the client, right?

cmurphy:
Good point, my brain had it backwards >.< changed it to `--cert-file` and `--key-file`

flavio (changes requested):
That looks pretty good to me. The only objection I have is about the amount of stuff we keep adding to the `main.rs` file. I think we should move all this part about starting the http(s) server to a dedicated module, and then just leverage a `pub(mod)` fuction/struct method from the `main`.

flavio (approved):
That's awesome, I think we can merge it.

Thanks @cmurphy !

ereslibre (approved):
LGTM, thank you @cmurphy! If we find a way to make some parts more idiomatic I think we can submit that later.

PR #17, opened 2021-01-29
Allow user to define multiple policies

flavio:
This PR allows policy server to load multiple policies. Each policy has also its own settings, which tune how they behave.

The policies configuration file is a yml file like the following one:

```yaml
valid:
  url: file:///home/flavio/hacking/kubernetes/chimera/policy-server/namespace-validate-policy.wasm
  settings:
    valid_namespace: valid
flavio:
  url: file:///home/flavio/hacking/kubernetes/chimera/policy-server/namespace-validate-policy.wasm
  settings:
    valid_namespace: flavio
empty:
  url: file:///home/flavio/hacking/kubernetes/chimera/policy-server/namespace-validate-policy.wasm
  settings:
    foo: bar
nothing:
  url: file:///home/flavio/hacking/kubernetes/chimera/policy-server/namespace-validate-policy.wasm
```

Each policy is identified by a unique ID (`valid`, `flavio`, `empty`, `nothing` in the example above), this ID is used to build the endpoint of the policy. All the policies are exposed by the web API under the `/validate/<policyID>` endpoint.

Given the previous example, the policy server would reply to `POST` requests sent to:

  * `/validate/valid`
  * `/validate/flavio`
  * `/validate/empty`
  * `/validate/empty`

Finally, as you can see the same policy can be loaded several times, each time with a different set of settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,99 @@
-Experimental rewrite in rust of the chimera admission controller.
+> **Note well:** don't forget to checkout [Chimera's documentation](https://chimera-kube.github.io/chimera-book/)
+> for more information
 
-Leverages waPC instead of WASI to write policies.
+# policy-server
+
+`policy-server` is a
+[Kubernetes dynamic admission controller](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
+that uses Chimera Policies to validate admission requests.
+
+Chimera Policies are simple [WebAssembly](https://webassembly.org/)
+modules.
+
+# Deployment
+
+We recommend to rely on the [chimera-controller](https://github.com/chimera-kube/chimera-controller)
+and the [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
+provided by it to deploy the Chimera stack.
+
+## Configuring policies
+
+A single instance of `policy-server` can load multiple Chimera policies. The list
+of policies to load, how to expose them and their runtime settings are handled
+through a policies file.
+
+By default `policy-server` will load the `policies.yml` file, unless the user
+provides a different value via the `--policies` flag.
+
+This is an example of the policies file:
+
+```yml
+dedicated_nodes_tenant_A:
+  url: registry://ghcr.io/chimera-kube/policies/pod-toleration:v0.0.2
+  settings:
+    taint_key: dedicated
+    taint_value: tenantA
+    allowed_groups:
+      - administrators
+      - tenant-a-users
+dedicated_nodes_tenant_B:
+  url: registry://ghcr.io/chimera-kube/policies/pod-toleration:v0.0.2
+  settings:
+    taint_key: dedicated
+    taint_value: tenantB
+    allowed_groups:
+      - administrators
+      - tenant-b-users
+namespace_simple:
+  url: file:///tmp/namespace-validate-policy.wasm
+  settings:
+    valid_namespace: chimera-approved
+```
+
+The YAML file contains a dictionary with strings as keys, and policy objects as values.
+
+The key that identifies a policy is used by `policy-server` to expose the policy
+through its web interface. Policies are exposed under `/validate/<policy id>.
+
+For example, given the configuration file from above, the following API endpoint
+would be created:
+
+  * `/validate/dedicated_nodes_tenant_A`: this exposes the `pod-toleration:v0.0.2`
+    policy. The Wasm module is downloaded from the OCI registry of GitHub.
+  * `/validate/dedicated_nodes_tenant_B`: this exposes the `pod-toleration:v0.0.2`
+    policy. The Wasm module is downloaded from the OCI registry of GitHub.
+  * `/validate/namespace_simple`: this exposes the `namespace-validate-policy`
+    policy. The Wasm module is loaded from a local file located under `/tmp/namespace-validate-policy.wasm`.
+
+It's common for policies to allow users to tune their behaviour via ad-hoc settings.
+These customization parameters are provided via the `settings` dictionary.
+
+For example, given the configuration file from above, the `namespace_simple` policy
+will be invoked with the `valid_namespace` parameter set to `chimera-approved`.
+
+Note well: it's possible to expose the same policy multiple times, each time with
+a different set of paraments. For example, given the configuration file from above,
+this is done by the `dedicated_nodes_tenant_A` and by the `dedicated_nodes_tenant_B` endpoints.
+
+The Wasm file providing the Chimera Policy can be either loaded from
+the local filesystem or it can be fetched from a remote location. The behaviour
+depends on the URL format provided by the user:
+
+* `file:///some/local/program.wasm`: load the policy from the local filesystem
+* `https://some-host.com/some/remote/program.wasm`: download the policy from the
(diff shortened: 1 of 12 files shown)


src/wasm_fetcher/mod.rs
```diff
@@ -35,3 +35,15 @@ pub(crate) fn parse_wasm_url(
     _ => Err(anyhow!("unknown scheme: {}", parsed_url.scheme())),
   }
 }
+
+pub(crate) async fn fetch_wasm_module(url: String) -> Result<String> {
+  let fetcher = match parse_wasm_url(
+    &url, false, //TODO: remote insecure
+    false, //TODO: remote non tls
```
cmurphy, line 42: :eyes:

src/api.rs
```diff
@@ -9,17 +9,27 @@ pub(crate) async fn route(
     req: hyper::Request<hyper::Body>,
     tx: mpsc::Sender<wasm::EvalRequest>,
 ) -> Result<hyper::Response<hyper::Body>, hyper::Error> {
-    match (req.method(), req.uri().path()) {
-        (&Method::POST, "/validate") => handle_post_validate(req, tx).await,
+    match req.method() {
+        &Method::POST => {
+            let path = String::from(req.uri().path());
+            if path.starts_with("/validate/") {
+                handle_post_validate(req, path.trim_start_matches("/validate/").to_string(), tx)
```
cmurphy, line 16: With there being separate endpoints for each policy, does this mean that the chimera controller will be creating a separate webhookconfig for every policy pointing to each `/validate/<id>` endpoint? Then kubernetes will have to make a round trip for every policy on every pod request. Why do we not just point the webhook to `/validate` and iterate over every configured policy on this end?

README.md
```diff
@@ -1,3 +1,96 @@
...
+
+We recommend to rely on the [chimera-controller](https://github.com/chimera-kube/chimera-controller)
+and the [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
+provided by it to deploy the Chimera stack.
+
+## Configuring policies
+
+A single instance of `policy-server` can load multiple Chimera policies. The list
+of policies to load, how to expose them and their runtime settings are handled
+through a `policies.yml` file.
```
cmurphy, line 23: The `--policies` flag is required and there's no default value, so would be nice to include an example of running the command with the flag.

cmurphy (commented):
Awesome work! I left some small comments inline.

One minor comment on the first commit message: it refers to a `policies.conf` file but then the readme refers to `policies.yml`, and actually there's no default so it can be called whatever the user wants.

Something to think about for the future is it might be nice if the config file could somehow be validated against the policies server-side. For instance, if I forget to include the `settings` for a policy, I won't know about it until I make a request, and then the error message about the invalid config is exposed to the user.

README.md
```diff
@@ -1,3 +1,96 @@
...
+
+This is an example of the file:
+
+```yml
+dedicated_nodes_tenant_A:
+  url: registry://ghcr.io/chimera-kube/policies/pod-toleration:v0.0.2
+  settings:
+    taint_key: dedicated
+    taint-value: tenantA
```
ereslibre, line 32:
```suggestion
    taint_value: tenantA
```

policies.yml.example
```diff
@@ -0,0 +1,17 @@
+dedicated_nodes_tenant_A:
+  url: registry://ghcr.io/chimera-kube/policies/pod-toleration:v0.0.2
+  settings:
+    taint_key: dedicated
+    taint-value: tenantA
```
ereslibre, line 5:
```suggestion
    taint_value: tenantA
```

policies.yml.example
```diff
@@ -0,0 +1,17 @@
...
+    allowed_groups:
+      - administrators
+      - tenant-a-users
+dedicated_nodes_tenant_B:
+  url: registry://ghcr.io/chimera-kube/policies/pod-toleration:v0.0.2
+  settings:
+    taint_key: dedicated
+    taint-value: tenantB
```
ereslibre, line 13:
```suggestion
    taint_value: tenantB
```

README.md
```diff
@@ -1,3 +1,96 @@
...
+    allowed_groups:
+      - administrators
+      - tenant-a-users
+dedicated_nodes_tenant_B:
+  url: registry://ghcr.io/chimera-kube/policies/pod-toleration:v0.0.2
+  settings:
+    taint_key: dedicated
+    taint-value: tenantB
```
ereslibre, line 40:
```suggestion
    taint_value: tenantB
```

ereslibre (commented):
Besides @cmurphy and some changes I suggested here, I think it's worth merging this as it is. We can make the code more rusty later on, no need to block on that IMO -- it's fine for the time being.

src/policies.rs
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+// Reads the policies configuration file, returns a HashMap with String as value
+// and Policy as values. The key is the name of the policy as provided by the user
+// inside of the configuration file. This name is used to build the API path
+// exposing the policy.
+pub(crate) fn read_policies_file(path: &str) -> Result<HashMap<String, Policy>> {
+  let settings_file = File::open(path).unwrap();
+  let ps: HashMap<String, Policy> = serde_yaml::from_reader(&settings_file)?;
```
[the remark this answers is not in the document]
flavio: fixed

README.md
flavio: fixed

src/wasm_fetcher/mod.rs
flavio: Yes, this is something ugly. Once https://github.com/chimera-kube/policy-server/pull/16/ is merged this can go away, because the registry pulling option will be handled via the dockerfile.

@ereslibre : do I get it right?

src/api.rs
flavio: Yes, you are correct. The chimera controller will create multiple `ValidatingWebhookConfigurations` objects, one per policy that is defined by the user. The configuration objects will point to the same `policy-server` instance, just to different endpoints.

It's also true that Kubernetes will have to make a round trip for every policy that has to be validated, but:

  * Different policies could process different kind of Kubernetes resources, or could be interested only to a subset of actions. This is expressed inside of the `ValidatingWebhookConfigurations`, Kubernetes knows better which of the chimera policies are relevant for a certain incoming request.
  * Maybe a request has to be evaluated by `N` policies. The roundtrip happens only when all the `N-1` policies have been accepted. An earlier rejection halts the full chain of validation.
  * Default timeout for a webhook call is 10 seconds for webhooks registered created using `admissionregistration.k8s.io/v1` (I took that from [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)), but this is tunable by the user. If we iterate over all the policies we might exceed that timeout.

Overall, I would not want `policy-server` to deal with all the nuances of iterating over all the policies. Kubernetes has already code for it, it's also a supported use case to have multiple webhook endpoints registered... So, I would start with something simple, we can always change that later on if needed

cmurphy (approved):
fixup commits could be squashed but otherwise lgtm

flavio:
Thanks @cmurphy!

Tomorrow I'll squash some of the commits and merge it

PR #22, opened 2021-02-08
Do not start the web API until the workers are ready

flavio:
Wait for all the workers to be ready before we start serving the web API.

Bootstraping the workers can take some time, depending on their number and the amount of policies each one of them has to initialize.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,10 @@
 use clap::{App, Arg};
-use std::{net::SocketAddr, process, thread};
+use std::{
+    net::SocketAddr,
+    process,
+    sync::{Arc, Barrier},
+    thread,
+};
 use tokio::{runtime::Runtime, sync::mpsc::channel};
 
 mod admission_review;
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -130,15 +135,20 @@ fn main() {
     }
 
     let (api_tx, api_rx) = channel::<EvalRequest>(32);
+    let pool_size = matches.value_of("workers").map_or_else(num_cpus::get, |v| {
+        usize::from_str_radix(v, 10).expect("error converting the number of workers")
+    });
+
+    let barrier = Arc::new(Barrier::new(pool_size + 1));
```
ereslibre, line 142: Didn't know barrier, nice!

src/worker.rs
```diff
@@ -56,16 +61,20 @@ impl WorkerPool {
...
     let mut tx_chans = Vec::<Sender<EvalRequest>>::new();
 
     for n in 1..=size {
       let (tx, rx) = channel::<EvalRequest>(32);
       tx_chans.push(tx);
       let ps = policies.clone();
+      let c = barrier.clone();
 
       thread::spawn(move || {
+        println!("spawining worker {}", n);
```
ereslibre, line 75:
```suggestion
        println!("spawning worker {}", n);
```

ereslibre (approved):
LGTM, thanks @flavio

flavio:
Thanks @ereslibre , I accepted your fix and then I squashed the commits together

PR #25, opened 2021-02-15
Allow user to choose where to download policies

flavio:
Prior to this commit the policies would be downloaded into the local execution  directory. This commit allows the user to specify where the policies files downloaded from a HTTP(s) server/OCI registry should be saved.

This is needed to run policy-server containerized. Without this patch the server would try to download the policies to `/usr/local/bin`, which is a bad thing; plus it's not writable by the unprivileged user running inside of the container.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -74,6 +74,13 @@ fn main() {
                 ),
         )
         .arg(
+            Arg::with_name("policies-download-dir")
+                .long("policies-download-dir")
+                .default_value(".")
+                .env("CHIMERA_POLICIES_DOWNLOAD_DIR")
+                .help("Download path for the policies"),
+        )
+        .arg(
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -119,9 +126,11 @@ fn main() {
         }
     };
 
+    let policies_download_dir = matches.value_of("policies-download-dir").unwrap();
```
ereslibre, line 129: Super nit: I wonder what happens if the command gets executed with `--poilcies-download-dir ""`, is anything preventing this from happening?

ereslibre (approved):
Thanks Flavio, LGTM and a comment.

src/main.rs
flavio: I've updated the code to ensure this doesn't happen.

PR #34, opened 2021-03-24
Mutating policies & code reorganizatoin

flavio:
This PR addresses the following issues:

* Fix a corner case when parsing `policies.yaml` file
* Reorganize the code base to simplify the process of sharing parts of its code
* Move the `chimera-policy-testdrive` project inside of this repository. This simplifies the process of sharing code between policy-server and policy-testdrive. Having both the projects live inside of the same repository allows to use the `cargo workspace` which greatly simplifies code reuse

New feature:
* Support mutation policies


crates/policy-testdrive/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
ereslibre, line 1: Is this LICENSE mandatory here or can be assumed inheriting the root one?

ereslibre (approved):

ereslibre:
LGTM besides the clippy complaints and the panic on the test.

crates/policy-testdrive/LICENSE
flavio: Probably not, I'll drop it

PR #35, opened 2021-03-25
Settings validation

flavio:
Leverage the `validate_settings` waPC function exposed by policies to ensure the configuration given by the user is valid.

# Shared behaviour

Policies that do not expose a `validate_settings` function are still loaded and used, though we can't ensure the user policy is correct.

# Policy Server

The policy server now exits at boot time with an error if one of its policies has a configuration error.

# policy-testdrive

Now validates settings before using the policy to validate the given input. If the settings are invalid it will exit with an error.


flavio:
Implements https://github.com/chimera-kube/policy-server/issues/30

crates/policy-evaluator/src/policy_evaluator.rs
```diff
@@ -105,4 +106,50 @@ impl PolicyEvaluator {
...
+                    .map_err(|e| anyhow!("cannot convert response: {:?}", e));
+                vr.unwrap_or_else(|e| SettingsValidationResponse {
+                    valid: false,
+                    message: Some(format!("error: {:?}", e)),
+                })
+            }
+            Err(err) => {
+                if let wapc::errors::ErrorKind::GuestCallFailure(m) = err.kind() {
+                    // Unfortunately waPC doesn't define a dedicated error
+                    if m.contains("No handler registered") {
```
ereslibre, line 136: Given this message could change, could we check the Wasm exports instead in order to determine if there exists a `validate_settings` function?

ereslibre (approved):
Thank you @flavio! A small comment that can be discussed/addressed later on

crates/policy-evaluator/src/policy_evaluator.rs
flavio: Yes, that's right. I'll try to do that, I've already played with wasm disassemblers "via code"

flavio: As discussed, let's merge this PR because it does some code reorganization. I'll file a new issue to track this improvement

PR #36, opened 2021-03-25
GitHub action: release static binary of policy-testdrive

flavio:
Build the policy-testdrive binary in a staic way and add the final artifact to the release page


flavio:
mhhh, clippy is complaining about something that is not related with this PR. Let's fix that with another PR

ereslibre (approved):

PR #38, opened 2021-03-26
Implement proper logging

flavio:
Use the tokio/tracing crate to implement tracing.

The policy-server is now honoring its `--debug` flag (previously the value of this flag wasn't handled).

Moreover a new flag `--log-fmt` has been introduced. This allows to choose between two types of log format:
  * `text`: simple text
  * `json`: json output

Regardless of the output format, all the events are structured logs.

Fixes https://github.com/chimera-kube/policy-server/issues/3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,6 +34,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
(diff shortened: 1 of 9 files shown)


flavio:
# Some screenshots

## Info level - text output

This is the default BTW

info-text

## Info level - JSON output

info-json

## Debug level - text output

debug-text

ereslibre (approved):

PR #39, opened 2021-03-28
Implement cluster context awareness

ereslibre:
This changeset allows the policy server to retrieve a list of well
known resources (namespaces, ingresses and services) and keeps them
updated by listing every five seconds.

This changeset also implements a callback on the policy evaluator that
allows the SDK's to consume this list of resources as a regular
string, that they can unmarshal to provide to the native language the
specific SDK targets.

By implementing the cluster context awareness inside the policy
evaluator, the listing of resources that is done in a dedicated
thread, expands the `policy-server` and the `policy-testdrive` as
well. This allows the `policy-testdrive` to automatically fetch this
list of well-known resources when it starts, so it can also be used to
perform complex tests that require polling information from a
Kubernetes cluster.

Some things that can be improved:

* Generalize types instead of relaying on a list of well known type of
  API resources. By generalizing we allow to also consume custom
  resources or specific Kubernetes types that are important to
  certain policies.

* Do not fetch & poll any kind of resource; delegate first fetch until
  first request is made by a policy, then include this type of
  resource on a watch.

* As specified on the previous point, use watchers instead of active
  polling.


crates/policy-evaluator/src/cluster_context.rs
```diff
@@ -0,0 +1,96 @@
...
+use lazy_static::lazy_static;
+use std::thread;
+
+lazy_static! {
+    static ref CLUSTER_CONTEXT_INITIALIZED: Mutex<bool> = Mutex::new(false);
+    static ref CLUSTER_CONTEXT: ClusterContext = ClusterContext::default();
+}
+
+#[derive(Default)]
+pub struct ClusterContext {
```
flavio, line 21: Please add some inline documentation explaining:
  * What is the ClusterContext struct, what's its purpose
  * How the synchronization works: all the part about RW locks
  * What are the limitations, plans for the future. Basically what you already wrote inside of the PR body

crates/policy-evaluator/src/policy_evaluator.rs
```diff
@@ -14,31 +14,42 @@ use chimera_kube_policy_sdk::settings::SettingsValidationResponse;
...
-            + Send,
-    ) -> Result<PolicyEvaluator> {
+    pub fn new(wasm_file: String, settings: serde_yaml::Mapping) -> Result<PolicyEvaluator> {
         let mut f = File::open(&wasm_file)?;
         let mut buf = Vec::new();
         f.read_to_end(&mut buf)?;
 
+        if let Err(error) = ClusterContext::init() {
+            println!("non fatal error: could not initialize a cluster context due to error: {}; context sensitive functions will not return any information", error);
+        }
```
flavio, line 51: I don't think this will ever happen because the `init` code `panic!`(s) when an error happens.

src/main.rs
```diff
@@ -120,31 +122,31 @@ fn main() {
     {
         Ok(a) => a,
         Err(error) => {
-            return fatal_error(format!("Error parsing arguments: {}", error));
+            panic!("error parsing arguments: {}", error);
```
flavio, line 125: I don't know if we want to `panic!` instead of using this homemade function. With the latest "logging" changes I just merged, the `fatal_error` does structured logs container the error message. This would no longer happen with `panic!`.

Maybe we could replace `fatal_error` with a macro that does something like:

```rust
error!(...)
panic!(...)
```

src/main.rs
```diff
@@ -120,31 +122,31 @@ fn main() {
...
         }
     };
 
     let cert_file = String::from(matches.value_of("cert-file").unwrap());
     let key_file = String::from(matches.value_of("key-file").unwrap());
     if cert_file.is_empty() != key_file.is_empty() {
-        return fatal_error("Error parsing arguments: either both --cert-file and --key-file must be provided, or neither.".to_string());
-    }
+        panic!("error parsing arguments: either both --cert-file and --key-file must be provided, or neither.");
```
flavio, line 132: Same concern here

src/main.rs
```diff
@@ -120,31 +122,31 @@ fn main() {
...
+    };
 
     let rt = match Runtime::new() {
         Ok(r) => r,
         Err(error) => {
-            return fatal_error(format!("Error initializing tokio runtime: {}", error));
+            panic!("error initializing tokio runtime: {}", error);
```
flavio, line 138: Same concern here

src/main.rs
```diff
@@ -120,31 +122,31 @@ fn main() {
...
     };
 
     let policies_file = matches.value_of("policies").unwrap_or(".");
     let mut policies = match read_policies_file(policies_file) {
         Ok(policies) => policies,
         Err(err) => {
-            return fatal_error(format!(
-                "Error while loading policies from {}: {}",
+            panic!(
+                "error while loading policies from {}: {}",
```
flavio, line 147: Same concern here

src/main.rs
```diff
@@ -153,11 +155,7 @@ fn main() {
         .map(|sources_file| match read_sources_file(sources_file) {
             Ok(sources) => sources,
             Err(err) => {
-                fatal_error(format!(
-                    "Error while loading sources from {}: {}",
-                    sources_file, err
-                ));
-                unreachable!();
+                panic!("error while loading sources from {}: {}", sources_file, err);
```
flavio, line 158: Same concern here

src/main.rs
```diff
@@ -188,7 +186,7 @@ fn main() {
         )) {
             Ok(path) => policy.wasm_module_path = path,
             Err(e) => {
-                return fatal_error(format!("Error while fetching policy {}: {}", policy.url, e));
+                panic!("error while fetching policy {}: {}", policy.url, e);
```
flavio, line 189: Same concern here

src/main.rs
```diff
@@ -220,7 +218,7 @@ fn main() {
     // at boot time
     main_barrier.wait();
     if !main_boot_canary.load(Ordering::SeqCst) {
-        return fatal_error(String::from("Couldn't init one of the workers"));
+        panic!("could not init one of the workers");
```
flavio, line 221: Same concern here

src/main.rs
```diff
@@ -231,18 +229,13 @@ fn main() {
         match server::new_tls_acceptor(&cert_file, &key_file) {
             Ok(t) => Some(t),
             Err(e) => {
-                return fatal_error(format!("Error while creating tls acceptor: {:?}", e));
+                panic!("error while creating tls acceptor: {:?}", e);
```
flavio, line 232: Same concern here

src/main.rs
```diff
@@ -231,18 +229,13 @@ fn main() {
...
             }
         }
     };
     rt.block_on(server::run_server(&addr, tls_acceptor, api_tx));
 
     if let Err(e) = wasm_thread.join() {
-        return fatal_error(format!("Error while waiting for worker threads: {:?}", e));
+        panic!("error while waiting for worker threads: {:?}", e);
```
flavio, line 239: Same concern here

flavio (changes requested):
Overall looks good to me, kudos!

There are just few changes we need

crates/policy-evaluator/src/policy_evaluator.rs
ereslibre: Yes, this is a leftover and I'm still divided on how to best handle this case. It's tricky because something could go wrong when initializing (e.g. could not find any usable configuration to build a Kube client -- no endless loop) vs "everything went fine" meaning we are in an endless loop polling for certain resources. I think I'll just remove this if and leave the inner panic/fatal_error.

ereslibre: It's also worth mentioning that a failure to create a Kube client should not be fatal and the policy-server/testdrive can *still* serve or load policies. The only issue is that if they try to read contextual resources they will be empty, but if they are "offline" checks they will work perfectly.

flavio: I'm all for exiting with an error if it's not possible to create a k8s client. If possible, I would however prefer to raise an error and then handle the exit inside of the `main`.

ereslibre: If we exit with an error then there's no possibility to run offline policies, do we want that? Also, I don't think there's a way to exit inside of the main -- we'd have to join the thread, but if everything went fine, the thread we are joining is an endless loop... Am I missing anything?

flavio: I think we have to work on defining a story around this type of failure, I have too many questions:
  * Should policy-server exit immediately with a failure when the client can't be created?
  * Should instead policy-server keep running, print a warning message and just serve empty results about `Namespace`, `Service`,... whatever?
  * Should policy-server look at the policies it has to host and act differently if none of the policies it loaded are "context aware" ones?

:exploding_head: too many variables... I can go on with other possibilities.

I would say, let's keep things simple and just panic

src/main.rs
ereslibre: Yes, going back to the home made function for now.

crates/policy-evaluator/src/cluster_context.rs
```diff
@@ -0,0 +1,126 @@
...
+// 2. Do not use polling. Perform an initial list request of resources
+// when the first request is performed by a guest. Then, cache and
+// watch in order to maintain an updated list: this significantly
+// reduces the data that needs to be pulled in a scheduled basis.
+//
+// Synchronization of the ClusterContext data structure happens by
+// protecting all inner shared resources with a RwLock. This enables
+// us to differentiate between reader locks and writer lock. Only one
+// writer exists, that will currently perform an active poll and
+// request information to the API server. Many writers may exist (many
```
flavio, line 46: Typo: `s/Many writers/Many readers/`

flavio (changes requested):
LGTM, just a small mistake inside of the inline documentation

crates/policy-evaluator/src/cluster_context.rs
ereslibre: LOL, thanks...

flavio (approved):
LGTM :rocket:

PR #44, opened 2021-04-01
Use the Kubewarden SDK crate we just published

flavio:
Just use the crate we published on crates.io, instead of using a random commit from git


ereslibre (approved):
Nice :)

Issue #50, opened 2021-04-07
policy-testdrive: remove confusing panic message


flavio:
Since the introduction of context-aware policies, the policy-testdrive can produce confusing message when there's no connection available to the k8s cluster defined inside of `~/.kube/config`.

For example, this is the output produced on my machine when my `minikube` server is down:

```shell
policy-testdrive --policy module.wasm --request-file pod-req-no-specific-apparmor-profile.json 
thread '<unnamed>' panicked at 'could not initialize Kubernetes client: Error loading kubeconfig: Failed to infer config.. cluster env: (Error loading kubeconfig: Unable to load in cluster config, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined), kubeconfig: (Error loading kubeconfig: Unable to load current context: )', crates/policy-evaluator/src/cluster_context.rs:74:31
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Settings validation result: SettingsValidationResponse { valid: true, message: None }
Policy evaluation results:
ValidationResponse { uid: "", allowed: true, patch_type: None, patch: None, status: None }
```

# Acceptance criteria

No panic message is printed, instead a nicer warning message should be shown. Something that tells:
  1. cannot reach the Kubernetes cluster defined inside of your configuration
  1. context-aware policies are not going to be working

ereslibre:
To add more information to this while reproducing. There are different behaviors that can happen with invalid configurations.

- A syntactically valid `kubeconfig` file, with unreachable Kubernetes cluster does not panic but does not inform the user that contextual policies might not work as expected.

- A syntactically invalid `kubeconfig` file does not panic, but does not inform the user.

- A completely missing `kubeconfig` file makes the `testdrive` panic.

Handle all these cases so we inform the user about the situation WRT to contextual policies as the acceptance criteria describes.,

PR #51, opened 2021-04-09
Fix threaded panic when a Kubernetes client cannot be built

ereslibre:
Whenever the `policy-server` or the `policy-testdrive` binaries try to
initialize a Kubernetes threaded update loop and fail, a panic message
was printed in the standard error.

Print a less critical message, and show a regular warning instead.

Fixes https://github.com/kubewarden/policy-server/issues/50


flavio (approved):
LGTM, I just left a minor change

PR #54, opened 2021-04-15
Update to latest version of kubewarden's client SDK

flavio:
This is needed to support policies written with v0.2.0 of the SDK

Once this is merged I'll tag a new release of policy-server

### DIFF
--- a/crates/policy-evaluator/src/validation_response.rs
+++ b/crates/policy-evaluator/src/validation_response.rs
@@ -55,9 +55,7 @@ impl ValidationResponse {
         pol_val_resp: &PolicyValidationResponse,
     ) -> Result<ValidationResponse> {
         let patch = match pol_val_resp.mutated_object.clone() {
-            Some(mut_obj_str) => {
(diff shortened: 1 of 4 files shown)


flavio:
mhhh... odd, the tests were not failing on my computer, I'll fix that

crates/policy-evaluator/src/validation_response.rs
```diff
@@ -167,7 +165,7 @@ mod tests {
             accepted: true,
             message: None,
             code: None,
-            mutated_object: Some(serde_json::to_string(&req_obj).unwrap()),
+            mutated_object: Some(req_obj.clone()),
```
ereslibre, line 168:
```suggestion
            mutated_object: Some(req_obj),
```

ereslibre (changes requested):
We can get rid of that clone. Apart from that, LGTM, thank you!

ereslibre (approved):
My bad on the previous review, got tricked by missing `-workspace` in `cargo test `, as you pointed out before. LGTM!

Issue #47, opened 2021-04-06
Ensure policy-testdrive binary is released when a new version is tagged


flavio:
As per subject, ensure the GitHub Action we have in place is actually publishing the `policy-testdrive` static linux binary on the `Releases` page.

This is important because it allows users to get started with kubewarden policies.

flavio:
Fixed

PR #55, opened 2021-04-15
Fix the release pipeline of `policy-testdrive`

flavio:
Fix the release action of policy-testdrive by ensuring the linux binary is properly compressed.
    
This fixes issue #47


ereslibre (approved):
LGTM, thank you. I think we should start getting used to use the supported prefixes, so the changelog will be even nicer (https://github.com/kubewarden/policy-server/blob/5a4a2bf39b9a624afe7bb33a14df1c81f608b6fc/.chglog/config.yml#L10-L13)

PR #56, opened 2021-04-15
refactor: split internal crates

ereslibre:
This is an initial refactor that splits a policy-fetcher crate out of
the policy-server.

Further refactors will follow reusing this logic.


flavio (approved):
LGTM, go on with that :rocket:

PR #58, opened 2021-04-30
refactor: due to dependency refactor

ereslibre:


Cargo.toml
```diff
@@ -10,8 +10,8 @@ edition = "2018"
 [dependencies]
 anyhow = "1.0"
 async-stream = "0.3.0"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.1.0" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.1.0" }
+policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", branch = "update-dependencies" }
+policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", branch = "refactor" }
```
ereslibre, line 14: This is not intended to be merged, this PR depends on the following PR's:

- https://github.com/kubewarden/policy-evaluator/pull/2
- https://github.com/kubewarden/policy-fetcher/pull/2

src/main.rs
```diff
@@ -98,13 +98,6 @@ fn main() {
                     their settings",
                 ),
         )
-        .arg(
-            Arg::with_name("policies-download-dir")
-                .long("policies-download-dir")
-                .default_value(".")
-                .env("KUBEWARDEN_POLICIES_DOWNLOAD_DIR")
-                .help("Download path for the policies"),
-        )
```
flavio, line 107: We must retain the ability to set the directory to be used as root for the downloads. This is important when running containerized, because the container user doesn't have a home directory at all:

https://github.com/kubewarden/policy-server/blob/851b8a8971c1e8d3e0063f7cb985245041440955/Dockerfile#L18-L25

flavio (changes requested):
I would like to have the cli flag back

PR #63, opened 2021-05-25
Update to latest version of policy-evaluator

flavio:
Refactoring of the policy-server code caused by changes done inside of policy-evaluator.


ereslibre (approved):
LGTM

PR #64, opened 2021-05-31
feat: use distroless build

ereslibre:
*WIP*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,25 +6,11 @@ COPY . .
 RUN cargo install --path .
 
 # final image
-FROM debian:buster-slim
+FROM gcr.io/distroless/cc:nonroot
 LABEL org.opencontainers.image.source https://github.com/kubewarden/policy-server
 
-COPY --from=builder /usr/local/cargo/bin/policy-server /usr/local/bin/policy-server
-
-RUN apt-get update && \
-    apt-get install -y ca-certificates && \
-    rm -rf /var/lib/apt/lists/*
-
-RUN adduser \
-  --disabled-password \
-  --gecos "" \
-  --no-create-home \
-  --home "/none" \
-  --shell "/sbin/nologin" \
-  --uid 2000 \
-  kubewarden
-USER kubewarden
+COPY --from=builder /usr/local/cargo/bin/policy-server /
 
 EXPOSE 3000
 
-ENTRYPOINT ["/usr/local/bin/policy-server"]
+CMD ["./policy-server"]


ereslibre:
*WIP*: checking several options for CA certificates. These are:

* Keep using `native-tls` (and thus, `openssl`)
  * Install CA certificates as [documented here](https://github.com/GoogleContainerTools/distroless/tree/main/cacerts) in the system CA store root
    * Includes the [`ca-certificates`](https://github.com/GoogleContainerTools/distroless/blob/2cac07c7c62b2dc52c4274a1810939fb06039ab9/cacerts/cacerts.bzl#L21-L27) from the Debian package.
* Use [`webpki-roots`](https://github.com/rustls/webpki-roots)
  * Includes the mozilla trusted roots

ereslibre:
I'll explore the "install" CA certificates on the base image first, instead of using `webpki-roots`. Making the different dependencies use `webpki-roots` is a bit tricky, although possible, and more importantly, a user might want to easily modify the set of CA certificates. Despite this should not be necessary (they can provide explicit CA chains to the policy-server and kwctl), they are in control of the last layers of the policy-server image if they want, what is notably easier than having to deal with how the crate dependency tree works, and how they can take certificates out from there, or add any kind of exceptions.

flavio:
Modifying the set of certificates that are part of a container image is ugly. You have to do that at build-time, at runtime this requires 1) root privileges to run `update-ca-certs` 2) a bash script that invokes `update-ca-certificate` when needed.

This is ugly, the right way to add extra certificates at runtime is by having the application deal with the dynamic load of certificates from the disk. This is something policy-server can already do.

Whatever you plan to use, run the image against trivy to get a feeling of the security issues it has. I ran trivy against the distroless/cc image and I was really disappointed:

```console
$ trivy i gcr.io/distroless/cc
2021-06-01T17:17:36.340+0200	INFO	Need to update DB
2021-06-01T17:17:36.340+0200	INFO	Downloading DB...
21.59 MiB / 21.59 MiB [---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 100.00% 24.70 MiB p/s 1s
2021-06-01T17:17:39.333+0200	INFO	Detected OS: debian
2021-06-01T17:17:39.333+0200	INFO	Detecting Debian vulnerabilities...
2021-06-01T17:17:39.340+0200	INFO	Number of PL dependency files: 0


gcr.io/distroless/cc (debian 10.9)
==================================
Total: 30 (UNKNOWN: 0, LOW: 17, MEDIUM: 4, HIGH: 9, CRITICAL: 0)
```

ereslibre:
> Modifying the set of certificates that are part of a container image is ugly. You have to do that at build-time, at runtime this requires 1) root privileges to run update-ca-certs 2) a bash script that invokes update-ca-certificate when needed.

This was never proposed to be done on runtime. Always on build time: the two options I presented were 1) use the "vendored" CA store from `webpki-roots` and not use the system CA store, 2) use the system CA store.

ereslibre:
I'll check with `trivy`, thanks for the tip!

ereslibre:
Closing for now, will open when I have something more specific.

PR #69, opened 2021-06-08
Bump dependencies in security report

ereslibre:
Fixes: https://github.com/kubewarden/policy-server/issues/67
Fixes: https://github.com/kubewarden/policy-server/issues/68


flavio:
Thanks!

I'm a bit surprised to see the security check fail again. According to [that](https://github.com/actions-rs/audit-check#limitations) the check should fail only for PRs that are coming from forked repositories; which is not the case here...

PR #75, opened 2021-07-12
Distributed tracing

flavio:
This PR focuses on the topics of generating proper events messages and sending them to distributed tracing platforms.

# Overview of the changes

* Better logging: fix tracing instrumentation inside of our codebase
* Drop the `--debug` cli flag, instead introduce a `--log-level` flag that allows the user to set the log level to `trace`, `debug`, `info`, `warn` and `error`
* Add two new logging "formats": Jaeger and [OpenTelemetry collector](https://opentelemetry.io/docs/collector/)
* Big refactoring of the `main.rs`:
  * Move some code from the `main` function to dedicated smaller functions, move some of these functions to separate files
  * Separate the initialization of the synchronous code and the asynchronous one in two distinctive sections. Before everything was kinda mixed

This "logging cleanup" has been made possible by sending trace events to [jaeger](https://www.jaegertracing.io/) and then doing some trial and error.

This a screenshot showing the trace events collected while evaluating one policy
image


ereslibre (approved):
As discussed, LGTM. There are some details that can be discussed later and with more knowledge about how some things plug together, like async runtimes, endless loops, etc...

Issue #76, opened 2021-07-12
Handle OPA/Gatekeeper policies


flavio:
This card is part of https://github.com/kubewarden/policy-evaluator/issues/14.

Policy server should be able to load and evaluate Wasm modules that have been originated by `opa build` and have been then annotated via kwctl (see https://github.com/kubewarden/kwctl/issues/55).

# Admission criteria

* policy server will load Rego-based policies that have been annotated by kwctl
* A Rego-based policy that has NOT been annotated by kwctl will not be loaded, and will cause the server to exit with an error
* policy server will prepare the right values expected by OPA and Gatekeeper policies

flavio:
To be more specific:

* policy-server requires all the wasm policies to be annotated via `kwctl annotate`.  This is the current behavior and we don't intend to change it
* If the kubewarden metadata includes information about the runtime to use: we will use this information to pick the right runtime (kubewarden/opa/gatekeeper)
* If the kubewarden metadata does NOT include information about the runtime to use (such as all the policies that currently exist), then we will assume the policy is a kubewarden one, and use that runtime

PR #82, opened 2021-08-18
opa: read policy execution mode when setting up the policy evaluator

ereslibre:
Fixes: https://github.com/kubewarden/policy-server/issues/76

When reading the policy and constructing a policy evaluator for it,
read the runtime annotation on the policy metadata and honor it. If
this annotation is present, fallback to the default settings of the
runtime: currently the Kubewarden Wapc backend.


flavio (approved):
LGTM

PR #85, opened 2021-09-13
Fix opentelemetry otlp integration

flavio:
This PR fixes the integration with opentelemetry otlp collector. When we updated the project dependencies we introduced 2 regressions:

1. The name of the service was not propagated by the otlp collector to other endpoints/collector
2. opentelemetry was not sending traces in batch mode. This is the recommended way for production deployments, otherwise the performance of the whole application will be affected by trace propagation

Finally, the PR adds a sample configuration file of otlp collector receiving events from policy-server and then sending them to a jaeger endpoint


ereslibre (approved):
Thanks 🙏

PR #87, opened 2021-09-16
report supported OPA builtins when `--version` is requested

ereslibre:


ereslibre:
Output:

```
~/projects/kubewarden/policy-server builtin-report » cargo run -- --version                                                                                                          k3d-k3s-default - ereslibre@desktop
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/policy-server --version`
policy-server 0.1.7

Open Policy Agent/Gatekeeper implemented builtins:
  - base64url.encode_no_pad
  - glob.quote_meta
  - hex.decode
  - hex.encode
  - json.is_valid
  - regex.find_n
  - regex.split
  - regex.template_match
  - semver.compare
  - semver.is_valid
  - sprintf
  - trace
  - urlquery.decode
  - urlquery.decode_object
  - urlquery.encode
  - urlquery.encode_object
  - yaml.is_valid
  - yaml.marshal
  - yaml.unmarshal
```

src/cli.rs
```diff
@@ -1,6 +1,9 @@
 use crate::settings::{read_policies_file, Policy};
 use anyhow::{anyhow, Result};
 use clap::{crate_authors, crate_description, crate_name, crate_version, App, Arg};
+use itertools::Itertools;
```
flavio, line 4: Do we need itertools because we have to iterate inside of the macro below?

flavio (approved):
LGTM

src/cli.rs
ereslibre: No, itertools include a `sorted` function that we use so the output is stable: https://docs.rs/itertools/0.10.1/itertools/fn.sorted.html

PR #93, opened 2021-09-28
Tracing improvements

flavio:
Improve the tracing experience: add the hostname of the machine (or Pod) as part of the tags. This makes possible to find all the trace events generated by a certain deployment of policy-server.

Also, remove some unwanted debug statements from the stdout.


ereslibre (approved):
I didn't test it, but LGM. Thanks!

Issue #102, opened 2021-10-07
Metric: policy evaluation counter


flavio:
The goal of this metric is to allow an operator to understand how many times a certain policy is evaluated.

The policy provides also some labels, that can be used to filter the metric and provide really specific data.

## Metric name

`kubewarden_policy_evaluations_total`

## Metric type

Counter

## Use cases

* How many requests have been evaluated inside of the cluster during a certain time frame
* How many requests have been blocked during the last hours
* How many requests have been accepted during the last hours
* How many requests have been mutated during the last hours
* How many Pod Creation requests have been blocked during the last hours

## Labels

* `policy_name`: name of the policy
* `resource_name`: name of the resource that has been evaluated by the policy
* `resouce_kind`: kind of the resource that has been evaluated by the policy. Eg: Pod, Deployment, Service,...
* `resource_namespace`: name of the Namespace the resource being evaluated belonged to. Of course that applies only to namespaced resources
* `resource_request_operation`: type of the operation (CREATE, UPDATE, DELETE)
* `accepted`: boolean value - final outcome of the evaluation
* `mutated`: boolean value - set to true when the request is accepted and mutated
* `error_code`: integer value - the error code that was returned by the policy. This only applies if the request was rejected, and is an optional field in all cases.

##  Deliverables

* [x] Instrument the policy-server code
* [x] Provide examples of PromQL queries that can take advantage of this metric

ereslibre:
I think it should has one more label at least:

- `error_code`: the returned error code, if the request was rejected. Empty it not present.

Also, we might want to add `error_message` as well, but I'm not so sure about this one.

flavio:
@ereslibre I think it would be used to have some rate metrics, for example (I haven't verified it):

Tracking the per-minute rate (averaged over 30s) at which requests are accepted:
```promql
rate(kubewarden_policy_evaluations_total{accepted="true"}[30s] == 1)*60
```

Issue #95, opened 2021-10-01
Expose policy metrics via a prometheus endpoint


flavio:
The goal is to have each PolicyServer instance expose metrics about the policies currently loaded.

The metrics should be exported as a prometheus endpoint.

TODO:
  * Define the metrics to be exposed
  * Understand the implications of exposing metrics via OpenTelemetry vs using a prometheus library

PR #108, opened 2021-10-15
Implement policy evaluation counter metric

ereslibre:
Fixes: kubewarden/policy-server#95
Fixes: kubewarden/policy-server#102

Besides the metric implementation, also:

- Update documentation
  - Update OpenTelemetry sample configuration
  - Add Prometheus sample configuration
  - Update DEVELOPMENT.md instructions
  - Remove unsupported `insecure` flag for jaeger in the OpenTelemetry collector
- Rename `AdmissionReview` to `AdmissionRequest` were appropriate
- Skip serializing optional fields if they are `null`: without skipping the serialization, some fields were set to `null` explicitly, causing an issue when unserializing on the guest policy side, given the attribute was present, but set to `null`. This was breaking policy evaluations on the policy server.

### DIFF
--- a/src/admission_review.rs
+++ b/src/admission_review.rs
@@ -20,17 +20,27 @@ pub(crate) struct AdmissionRequest {
     pub uid: String,
     pub kind: GroupVersionKind,
     pub resource: GroupVersionResource,
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


ereslibre:
Added documentation on opentelemetry-rust examples: https://github.com/open-telemetry/opentelemetry-rust/pull/653

Issue #88, opened 2021-09-17
Report loaded and updated policies in the log


kkaempf:
When I update the `pod-privileged` policy (because of kubewarden/kubewarden.io#26), `helm` just reports
> clusteradmissionpolicy.policies.kubewarden.io/privileged-pods configured

So I looked at the `policy-server` logs (using [k9s](https://k9scli.io)) to get some more information but it only shows

> Sep 17 08:56:33.432  INFO policy_server: policies download download_dir="/tmp/" policies_count=1 status="init"                                                                             
> pulling policy...
> Sep 17 08:56:35.216  INFO policy_server: policies download status="done" 
> Sep 17 08:56:35.216  INFO policy_server: kubernetes poller bootstrap status="init"
> Sep 17 08:56:35.216  INFO policy_server: kubernetes poller bootstrap status="done" 
> Sep 17 08:56:35.216  INFO policy_server: worker pool bootstrap status="init"
> Sep 17 08:56:35.216  INFO policy_server::kube_poller: spawning cluster context refresh loop                                                                                             
> Sep 17 08:56:35.229  INFO policy_server::worker_pool: spawning worker spawned=2 total=2
> Sep 17 08:56:35.230  INFO policy_server::worker_pool: spawning worker spawned=1 total=2                                                                                                    
> Sep 17 08:56:35.328  INFO policy_server: worker pool bootstrap status="done"                                                                                                               
> Sep 17 08:56:35.333  INFO policy_server::server: started HTTPS server address="0.0.0.0:8443"  

which doesn't tell me anything about

- which policy was pulled
- which version of this policy
- was it running before (is it an update) or not (is it a new policy)
- which other policies are running (in which version)

While the current log is helpful for developers, it doesn't contain information relevant for operators.

ereslibre:
Related PR's:

- https://github.com/kubewarden/policy-fetcher/pull/24
- https://github.com/kubewarden/policy-server/pull/119
- https://github.com/kubewarden/kwctl/pull/83

PR #119, opened 2021-11-04
feat: report loaded policies on startup

ereslibre:
Fixes: https://github.com/kubewarden/policy-server/issues/88

The reported information includes:

- The name that was given to the policy in the configuration
- The SHA-256 sum of the policy itself
- Whether the policy is mutating or not


jvanz (approved):
LGTM! :D

viccuad (approved):

PR #120, opened 2021-11-04
chore: pin base64ct to 1.1.1 version for Rust Edition 2018

ereslibre:
A newer version of base64ct (1.2.x) is bumped to Rust Edition
2021. Keep a pinned version to 1.1.1 of base64ct that did depend on
Rust Edition 2018.

We can get rid of this when we are ready to move to Edition 2021.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,8 +271,7 @@ checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 [[package]]
 name = "base64ct"
 version = "1.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jvanz (approved):
LGTM! :D

viccuad (approved):
Thanks!

~~I tried a `cargo fix; <bump edition in cargo.tml>; cargo test`, and it's all fine, but I suppose we need some manual testing for all features, just in case.~~

flavio:
> I tried a cargo fix --edition; <bump edition in cargo.tml>; cargo test, and it's all fine, but I suppose we need some manual testing for all features, just in case.

I would rather spend that testing time to ensure everything works with the new edition. My 2 cents :)

In the meantime, let's merge this PR

PR #121, opened 2021-11-15
chore: add cargo audit configuration

ereslibre:
By adding a cargo audit configuration we are able to ignore security
advisories that don't apply.

Related: https://github.com/kubewarden/policy-server/issues/110

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,11 @@
+[advisories]
+ignore = [
+  "RUSTSEC-2020-0071", # `time` localtime_r segfault -- https://rustsec.org/advisories/RUSTSEC-2020-0071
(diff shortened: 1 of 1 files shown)


viccuad (approved):
LGTM. Still, I'm afraid of forgetting about advisories this way. Would love for GHA to have status "warning", so at least we are reminded.. pity.

ereslibre:
In this instance I'll hold the merge until more approvals are granted, given I want to ensure we are on the same page regarding something as potentially sensitive as this.

flavio (approved):
I share the same concerns as @viccuad , but on the other hand I don't think we should keep getting bothered by these notifications.

Maybe I would add some comments next to the issues to explain why we decided to mute them

flavio (approved):

PR #122, opened 2021-11-16
Policy evaluation latency metric.

jvanz:
Adds a metric to allow the operator check how long a policy evaluation takes.

### DIFF
--- a/src/metrics/policy_evaluations_latency.rs
+++ b/src/metrics/policy_evaluations_latency.rs
@@ -1,0 +1,20 @@
+use crate::metrics::PolicyEvaluation;
+use lazy_static::lazy_static;
+use opentelemetry::{metrics::ValueRecorder, KeyValue};
+use std::convert::TryFrom;
+use std::time::Duration;
+
+lazy_static! {
+    static ref POLICY_EVALUATION_LATENCY: ValueRecorder<u64> =
+        opentelemetry::global::meter(super::METER_NAME)
+            .u64_value_recorder("kubewarden_policy_evaluation_latency_milliseconds")
+            .init();
+}
+
+pub fn record_policy_latency(latency: Duration, policy_evaluation: &PolicyEvaluation) {
+    let millis_latency = u64::try_from(latency.as_millis()).unwrap_or(u64::MAX);
+    POLICY_EVALUATION_LATENCY.record(
+        millis_latency,
+        &Into::<Vec<KeyValue>>::into(policy_evaluation),
+    );
+}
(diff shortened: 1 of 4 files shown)


ereslibre (approved):
LGTM

src/metrics/policy_evaluations_latency.rs
```diff
@@ -0,0 +1,18 @@
...
+lazy_static! {
+    static ref POLICY_EVALUATION_LATENCY: ValueRecorder<f64> =
+        opentelemetry::global::meter(super::METER_NAME)
+            .f64_value_recorder("kubewarden_policy_evaluation_latency_seconds")
+            .init();
+}
+
+pub fn record_policy_latency(latency: Duration, policy_evaluation: PolicyEvaluation) {
+    POLICY_EVALUATION_LATENCY.record(
+        latency.as_secs_f64(),
```
ereslibre, line 15: Just a comment: is secs a good resolution? Shouldn't we use ms instead?

ereslibre: And rename the metric ofc, I think we should aim for milliseconds overall.

flavio: yeah, +1 for using milliseconds

src/metrics/policy_evaluations_latency.rs
```diff
@@ -0,0 +1,18 @@
...
+use std::time::Duration;
+
+lazy_static! {
+    static ref POLICY_EVALUATION_LATENCY: ValueRecorder<f64> =
+        opentelemetry::global::meter(super::METER_NAME)
+            .f64_value_recorder("kubewarden_policy_evaluation_latency_seconds")
+            .init();
+}
+
+pub fn record_policy_latency(latency: Duration, policy_evaluation: PolicyEvaluation) {
```
flavio, line 13: Should we pass `policy_evaluation` by reference?

ereslibre: I think it might have an impact on the `Into<Vec<KeyValue>>` impl, but we can try to do the impl for &PolicyEvaluation though.

@jvanz can you check?

flavio: I don't want to over-optimize things, but given this code is called on each evaluation I was thinking about avoiding copies as much as possible

flavio (approved):
LGTM

ereslibre (approved):

jvanz:
Fixed code format with `cargo fmt`. After the CI finishes the checks, I'll merge the code.

PR #124, opened 2021-11-17
New Grafana panel using other Kubewarden metrics.

jvanz:
Adds three more panels using the latest metrics added in Kubewarden. The metrics are the policy counter and policy evaluation latency metrics.

### DIFF
--- a/kubewarden-dashboard.json
+++ b/kubewarden-dashboard.json
@@ -1,46 +1,4 @@
 {
-  "__inputs": [
-    {
-      "name": "DS_PROMETHEUS",
-      "label": "Prometheus",
-      "description": "",
-      "type": "datasource",
-      "pluginId": "prometheus",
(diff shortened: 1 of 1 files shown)


jvanz:
Screenshot from 2021-11-17 17-28-40

kubewarden-dashboard.json
```diff
@@ -65,8 +23,8 @@
   "fiscalYearStartMonth": 0,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": null,
```
kkaempf, line 68: What is `id` and why is 26 the right value ?

kkaempf (approved):
lgtm

kubewarden-dashboard.json
jvanz: I do not know. I believe Grafana did this change when I've exported the dashboard.

PR #126, opened 2021-11-23
Integrate the CallbackHandler

flavio:
Ensure policies can invoke asynchronous code on the host. This is implemented by using the new CallbackHandler provided by policy-evaluator.

This allows the usage of the waPC host call that fetches the digest of a container image

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,7 +4,9 @@ extern crate policy_evaluator;
 
 use anyhow::Result;
 use opentelemetry::global::shutdown_tracer_provider;
+use policy_evaluator::callback_handler::CallbackHandlerBuilder;
 use policy_evaluator::policy_metadata::Metadata;
+use std::path::PathBuf;
 use std::{process, thread};
 use tokio::{runtime::Runtime, sync::mpsc, sync::oneshot};
 use tracing::{debug, error, info};
@@ -21,8 +23,6 @@ mod worker;
 
 mod worker_pool;
 use worker_pool::WorkerPool;
-
-use std::path::PathBuf;
 
 mod communication;
 use communication::{EvalRequest, KubePollerBootRequest, WorkerPoolBootRequest};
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -237,6 +266,13 @@ fn main() -> Result<()> {
         }
         info!(status = "done", "kubernetes poller bootstrap");
 
+        // Spawn the tokio task used by the CallbackHandler
+        let callback_handle = tokio::spawn(async move {
+            info!(status = "init", "CallbackHandler task");
+            callback_handler.loop_eval().await;
```
ereslibre, line 272: `loop_eval()` does an infinite loop: https://github.com/kubewarden/policy-evaluator/pull/67/files#diff-26f1a863d124e9b03db590e7ab50f50548b71b6e91fb40d0c2e510565743fb8cR104-R131. As far as I understand, `tokio::spawn()` **could** run on the same thread (from the docs: "The spawned task may execute on the current thread, or it may be sent to a different thread to be executed").

if it is executed in the current thread I fear we might never make progress on futures if we are stuck on an endless loop in our side.

Maybe this comes from not fully understanding the runtime properties of Tokio & futures, could this happen?

ereslibre (approved):
LGTM. Just a question.

src/main.rs
viccuad: I have the same feeling here, thanks for spotting it Rafa. Maybe the blocking future should be in its own `thread::spawn`, or use `tokio::spawn_blocking`? (https://docs.rs/tokio/1.14.0/tokio/task/fn.spawn_blocking.html)

viccuad (approved):
LGTM module the raised issue. Yay callbacks!

src/main.rs
flavio: this is not a problem, I found the real root cause of the performance issue: https://github.com/kubewarden/policy-evaluator/pull/67#issuecomment-984471982

flavio:
I've rebased against the main branch

PR #127, opened 2021-11-24
Dashboard color standards and better aggregations.

jvanz:
During a past demo, I ended up performing some improvements in our Grafana Dashboard to enhance the data visualization. 

The Grafana Dashboard has been updated enforcing a color standard. Now all the graphics related to mutated request are purple, accepted requests are green, reject requests are red.

Furthermore, some queries have been updated to aggregate different time series in one single line.  This is necessary because we have many label in the metrics and each different value generate another time series. Thus, we can leave with many lines in the graph which does not tell use useful info.

Screenshot from 2021-11-24 15-33-04

Screenshot from 2021-11-24 15-33-41

Screenshot from 2021-11-24 15-34-06

Screenshot from 2021-11-24 15-35-09

### DIFF
--- a/kubewarden-dashboard.json
+++ b/kubewarden-dashboard.json
@@ -1,4 +1,46 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "8.2.3"
+    },
+    {
+      "type": "panel",
+      "id": "heatmap",
+      "name": "Heatmap",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    },
+    {
+      "type": "panel",
+      "id": "stat",
+      "name": "Stat",
+      "version": ""
+    },
+    {
+      "type": "panel",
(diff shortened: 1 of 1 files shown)


kubewarden-dashboard.json
```diff
@@ -1555,8 +1639,8 @@
       {
         "current": {
           "selected": false,
-          "text": "psp-user-group",
-          "value": "psp-user-group"
+          "text": "privileged-pods-policy",
+          "value": "privileged-pods-policy"
```
ereslibre, line 1643: I was wondering this on the previous version of the dashboard. Should we default the dashboard to a specific policy, or provide it "clean"?

kubewarden-dashboard.json
```diff
@@ -1584,5 +1668,5 @@
   "timezone": "",
   "title": "Kubewarden",
   "uid": "r3Pw-1O7z",
-  "version": 5
+  "version": 2
```
ereslibre, line 1671: I guess this is not an issue right now :)

ereslibre (commented):
LGTM, a couple of comments though.

viccuad (approved):
Nice separation of accepted/mutated, and colors. Apart from the questions raised, LGTM.

kubewarden-dashboard.json
jvanz: Yeah, an empty value is better. I just updated the dashboard removing the variable default value.

kubewarden-dashboard.json
jvanz: :+1:

jvanz:
As the requests from @ereslibre are small and I've already pushed them. I'm merging this. Thus, we can publish the dashboard in the Grafana.com

@ereslibre if you have more changes, let me know. I'll open another PR for them.

PR #130, opened 2021-11-30
 Run end-to-end tests daily basis.

### LINKED ISSUE #129: Run Policy Server end-to-end tests every day

jvanz:
Adds a new Github workflow to run the Kubewarden end-to-end test every day.

Closes #129

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,0 +1,16 @@
+name: Build container image every change.
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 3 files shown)


flavio:
Do we need that in two places (policy-server and kubewarden)? Can't we simply run it only in one place?

jvanz:
> Do we need that in two places (policy-server and kubewarden)? Can't we simply run it only in one place?

For now yes. Because we are not running it in the release process.

PR #131, opened 2021-12-01
replace openssl with rustls

ereslibre:


Cargo.toml
```diff
@@ -11,25 +11,25 @@ edition = "2018"
 anyhow = "1.0"
 async-stream = "0.3.0"
 itertools = "0.10.1"
-policy-evaluator = { git = "https://github.com/kubewarden/policy-evaluator", tag = "v0.2.4" }
-policy-fetcher = { git = "https://github.com/kubewarden/policy-fetcher", tag = "v0.1.19" }
+policy-evaluator = { git = "https://github.com/ereslibre/policy-evaluator", branch = "rustls-tls" }
+policy-fetcher = { git = "https://github.com/ereslibre/policy-fetcher", branch = "rustls-tls" }
```
jvanz, line 15: Do not we need to update the dependencies to the `kubewarden` org again? ;)

ereslibre: Yes, this is temporary, so it's easy to test and it can run on CI until the PR's over there merge.

flavio (approved):
LGTM, thanks for having done this change!

Dockerfile
```diff
@@ -32,12 +26,9 @@ LABEL org.opencontainers.image.source https://github.com/kubewarden/policy-serve
 
 USER root
 
-COPY --from=sle /etc/passwd /etc/passwd
-COPY --from=sle /*.rpm /
+COPY --from=builder /etc/passwd /etc/passwd
```
viccuad, line 29: I wonder if this is adding more users than needed, now that it happens from the openSUSE image.

flavio: To be honest, adding a user to the container can be done with a simple `echo` command:

```shell
# By default we will run as this user...
RUN echo "policy-server:x:65533:65533::/tmp:/sbin/nologin" >> /etc/passwd
# Add the default GID to /etc/group for completeness.
RUN echo "policy-server:x:65533:policy-server" >> /etc/group
```

Taken from [here](https://github.com/kubernetes/git-sync/blob/dc324f42839a6b2fb28f0305bc48c7b4dbb9fe0a/Dockerfile.in#L68-L75)

Dockerfile
```diff
@@ -8,37 +8,24 @@ RUN zypper in -y curl && \
...
+RUN echo "policy-server:x:65533:policy-server" >> /etc/group
 
-COPY --from=sle /etc/passwd /etc/passwd
-COPY --from=sle /*.rpm /
 COPY --from=builder /usr/local/cargo/bin/policy-server /usr/local/bin/policy-server
 
-RUN rpm --install /*.rpm; rm -f /*.rpm
-
-USER kubewarden
+USER 65533:65533
```
flavio, line 28: Nice touch, this makes clear the app inside of the container image isn't running as root. That's great for PSPs :+1:

flavio (approved):
LGTM

Dockerfile
ereslibre: Done, thank you!

ereslibre:
I just verified that this build does not crash in arm64. I built the container image and deployed it on an arm64 machine and it works as expected.

PR #134, opened 2021-12-20
Fix container image build

flavio:
Update the policy-evaluator crate to ensure openssl-sys is not pulled in.

This fixes container image building.


ereslibre (approved):
I think we are good to early merge here.

PR #138, opened 2021-12-23
cross-compile container images

ereslibre:
Create multi-arch container images by first cross compiling the
binaries for the different platforms, and then creating the container
images by copying the binaries.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,12 @@
+# Due to an issue with linking when cross-compiling, specify the
+# linker and archiver for cross-compiled targets.
+#
+# More information: https://github.com/rust-lang/cargo/issues/4133
+
+[target.x86_64-unknown-linux-musl]
+linker = "x86_64-linux-musl-ld"
+ar = "x86_64-linux-musl-ar"
+
+[target.aarch64-unknown-linux-musl]
+linker = "aarch64-linux-musl-ld"
+ar = "aarch64-linux-musl-ar"
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -1,57 +1,113 @@
+name: build container image
 on:
   push:
     branches:
     - main
     tags:
     - 'v*'
-
-name: build container image
-
 jobs:
-  build:
+  ci:
+    uses: .github/workflows/tests.yml@main
+  build-x86_64:
+    name: Build x86_64 binary
+    runs-on: ubuntu-latest
+    needs:
+      - ci
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - run: rustup target add x86_64-unknown-linux-musl
+      - name: Setup musl for x86_64
+        run: |
+          curl https://musl.cc/x86_64-linux-musl-cross.tgz | tar -xz
+          echo "$PWD/x86_64-linux-musl-cross/bin" >> $GITHUB_PATH
+      - name: Build policy-server
+        env:
+          CC: x86_64-linux-musl-gcc
+        run: cargo build --target=x86_64-unknown-linux-musl --release
+      - name: Upload policy-server
+        uses: actions/upload-artifact@v2
+        with:
+          name: policy-server-x86_64
+          path: target/x86_64-unknown-linux-musl/release/policy-server
+  build-aarch64:
+    name: Build aarch64 binary
+    runs-on: ubuntu-latest
+    needs:
+      - ci
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - run: rustup target add aarch64-unknown-linux-musl
+      - name: Setup musl for aarch64
+        run: |
+          curl https://musl.cc/aarch64-linux-musl-cross.tgz | tar -xz
+          echo "$PWD/aarch64-linux-musl-cross/bin" >> $GITHUB_PATH
+      - name: Build policy-server
+        env:
+          CC: aarch64-linux-musl-gcc
+        run: cargo build --target=aarch64-unknown-linux-musl --release
+      - name: Upload policy-server
+        uses: actions/upload-artifact@v2
+        with:
+          name: policy-server-aarch64
+          path: target/aarch64-unknown-linux-musl/release/policy-server
+  build-container-image:
     name: Build container image
     runs-on: ubuntu-latest
+    needs:
+     - build-x86_64
+     - build-aarch64
     steps:
-      -
-        name: Checkout code
+      - name: Checkout code
         uses: actions/checkout@v2
-      -
-        name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
-      -
-        name: Set up Docker Buildx
+      - name: policy-server-x86_64
+        uses: actions/download-artifact@v2
+        with:
+          name: policy-server-x86_64
+          path: /tmp/kubewarden
+      - name: policy-server-aarch64
+        uses: actions/download-artifact@v2
+        with:
+          name: policy-server-aarch64
+          path: /tmp/kubewarden
+      - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-      -
-        name: Login to GitHub Container Registry
+      - name: Login to GitHub Container Registry
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      -
-        name: Build and push development container image
+      - name: Build and push development container image
         if: ${{ startsWith(github.ref, 'refs/heads/') }}
         uses: docker/build-push-action@v2
         with:
-          context: .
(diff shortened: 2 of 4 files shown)


ereslibre:
Converting to draft because of the usage of `musl.cc`.

.github/workflows/container-image.yml
```diff
@@ -4,54 +4,106 @@ on:
...
+      - name: Build policy-server
+        env:
+          CC: x86_64-linux-musl-gcc
+        run: |
+          cargo build --target=x86_64-unknown-linux-musl --release
+      - name: Upload policy-server
+        uses: actions/upload-artifact@v2
+        with:
+          name: policy-server-x86_64
+          path: target/x86_64-unknown-linux-musl/release/policy-server
```
flavio, line 34: I guess this why you changed the `.Dockerignore` to allow the `target` directory. I would just copy this file (and the aarch64 one) into the root of the git checkout. Obviously using two different names (like `policy-server-x86_64`)

flavio (commented):
Cool stuff :clap:

viccuad (commented):
Thanks for this, this is great! I'm amazed that the compilation time went down to 15 minutes.

Did we talk about performance of running code already? I have read that Musl is not as efficient as libc on mem allocation. I suppose not a problem for our usage, but maybe worth checking.

With this, we will be moving our targets from Rust Tier 1 (guaranteed to work) to [Tier 2](https://doc.rust-lang.org/rustc/platform-support.html#tier-2-with-host-tools) (guaranteed to build). For x86_64 this means a reduction in assurances. I suppose it's not a problem, but something to also have in mind.

.cargo/config
```diff
@@ -0,0 +1,7 @@
+[target.x86_64-unknown-linux-musl]
+linker = "x86_64-linux-musl-ld"
+ar = "x86_64-linux-musl-ar"
+
+[target.aarch64-unknown-linux-musl]
+linker = "aarch64-linux-musl-ld"
+ar = "aarch64-linux-musl-ar"
```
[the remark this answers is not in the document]
ereslibre: Without it, `ring`'s `build.rs` (https://github.com/briansmith/ring/blob/main/build.rs) fails to cross-compile.

Related: https://github.com/rust-lang/cargo/issues/4133

Dockerfile
```diff
@@ -1,32 +1,21 @@
...
+COPY --from=common /etc/group /etc/group
+COPY --chmod=0755 policy-server-aarch64 /
 
+# common final steps
+FROM build-${TARGETARCH}
+USER 65533:65533
 EXPOSE 3000
-
-ENTRYPOINT ["/usr/local/bin/policy-server"]
+ENTRYPOINT ["/policy-server"]
```
viccuad, line 21: Silly question, is there a reason to put the bin out of path?

viccuad (approved):
LGTM! yay!
Do we want to cross-compile on development too?
I tried here `cargo build --target x86_64-unknown-linux-musl` and it failed on the ring issue, so from what I understand it's not using `.cargo/config`.

.cargo/config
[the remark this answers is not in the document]
flavio: Oh, that explains the issues I've seen with musl compilation with ring (even without cross build). Thanks

flavio (commented):
Overall great :clap: 

I left some comments here and there

flavio:
> Do we want to cross-compile on development too? I tried here `cargo build --target x86_64-unknown-linux-musl` and it failed on the ring issue, so from what I understand it's not using `.cargo/config`.

Actually that would be nice, just to be able to reproduce bugs locally...

.github/workflows/container-image.yml
```diff
@@ -4,54 +4,104 @@ on:
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - run: rustup target add x86_64-unknown-linux-musl
+      - name: Setup musl for x86_64
+        run: |
+          curl -O https://musl.cc/x86_64-linux-musl-cross.tgz
```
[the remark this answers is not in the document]
ereslibre: I also think this is a problem. I see two possible options:

1. Upload the tarball used for a release somewhere (along with the release?). This does not seem "pretty". Another issue is data contention on hosted platforms: for how long would that tarball be kept -- e.g. had we uploaded it to GitHub Actions? If we upload it as a release asset I think data contention should be take for granted, but it's weird to show final binaries and the libc along with them -- maybe we could somehow add the libc to the source code tarball that is part of the release?
2. Use a reproducible package manager on top, e.g. nix. Everyone on the team (and the CI) would always refer to the same versions of all dependencies.

I think it's not a blocker **right now**, but I completely see this as a blocker for a first stable and supported reelase.

Dockerfile
ereslibre: We can put it under any `bin` dir but I honestly don't see a strong reason to do so. The tree would be as @flavio described.

Dockerfile
```diff
@@ -1,32 +1,21 @@
...
 RUN echo "policy-server:x:65533:65533::/tmp:/sbin/nologin" >> /etc/passwd
-# Add the default GID to /etc/group for completeness.
 RUN echo "policy-server:x:65533:policy-server" >> /etc/group
 
-COPY --from=builder /usr/local/cargo/bin/policy-server /usr/local/bin/policy-server
+# amd64-specific
+FROM scratch AS build-amd64
+COPY --from=common /etc/passwd /etc/passwd
+COPY --from=common /etc/group /etc/group
+COPY --chmod=0755 policy-server-x86_64 /
```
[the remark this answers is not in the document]
ereslibre: Good catch. Fixed.

ereslibre:
>> Do we want to cross-compile on development too? I tried here cargo build --target x86_64-unknown-linux-musl and it failed on the ring issue, so from what I understand it's not using .cargo/config.

> Actually that would be nice, just to be able to reproduce bugs locally...

I also agree. I think is a good strategy. I would push for that right after this work. We can discuss about it. My preferred mechanism would be to have this as a nix project, so our dependencies are pinned and easily built (no need to fill pre-requirements prior to build other than having nix installed).

I think is worth checking, let me know if that's good with you and if you think we should plan for that soon.

ereslibre:
> Do we want to cross-compile on development too?

You can manually do this already, but is not pretty. Let me know if this fixes the build & link issue for you.

```
$ curl https://musl.cc/x86_64-linux-musl-cross.tgz | tar -xz
$ export PATH=$PWD/x86_64-linux-musl-cross/bin:$PATH
$ cargo build --target x86_64-unknown-linux-musl
```

Same for targeting aarch64 from x86_64:

```
$ curl https://musl.cc/aarch64-linux-musl-cross.tgz | tar -xz
$ export PATH=$PWD/aarch64-linux-musl-cross/bin:$PATH
$ cargo build --target aarch64-unknown-linux-musl
```

This should work AFAICT.

Dockerfile
viccuad: doh true, using scratch indeed. Don't mind, then :).

.github/workflows/container-image.yml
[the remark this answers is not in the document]
flavio: Yes, this is not a blocker right now. let's look into that later on

.github/workflows/container-image.yml
```diff
@@ -1,57 +1,102 @@
...
+        run: |
+          curl https://musl.cc/x86_64-linux-musl-cross.tgz | tar -xz
+          echo "$PWD/x86_64-linux-musl-cross/bin" >> $GITHUB_PATH
+      - run: touch policy-server-x86_64
+      # - name: Build policy-server
+      #   env:
+      #     CC: x86_64-linux-musl-gcc
+      #   run: |
+      #     cargo build --target=x86_64-unknown-linux-musl --release
+      #     mv target/x86_64-unknown-linux-musl/release/policy-server policy-server-x86_64
```
flavio, line 25: This smells like some debugging going on :)

.github/workflows/container-image.yml
```diff
@@ -1,57 +1,102 @@
...
+        run: |
+          curl https://musl.cc/aarch64-linux-musl-cross.tgz | tar -xz
+          echo "$PWD/aarch64-linux-musl-cross/bin" >> $GITHUB_PATH
+      - run: touch policy-server-aarch64
+      # - name: Build policy-server
+      #   env:
+      #     CC: aarch64-linux-musl-gcc
+      #   run: |
+      #     cargo build --target=aarch64-unknown-linux-musl --release
+      #     mv target/aarch64-unknown-linux-musl/release/policy-server policy-server-aarch64
```
flavio, line 52: Same here

flavio (commented):
I think there's still some work going on, am I right? This doesn't look ready to be merged

.github/workflows/container-image.yml
ereslibre: Yes, I am still checking that the image can be built properly on my fork. I will ping you here when the build has succeeded in Github actions in my fork.

ereslibre:
> I think there's still some work going on, am I right? This doesn't look ready to be merged

This is mergeable. I was testing GitHub behaviors when building the container image. Certain parts are a bit of trial and error. It's good now.

jvanz (approved):
LGTM! :D

flavio (approved):
Great, go ahead with it :clap:

Issue #137, opened 2021-12-22
Allow mutating policies to behave as validating


ereslibre:
Add a new setting for every policy that specifies whether this policy is mutating or not. This field would be added to the [policy settings](https://github.com/kubewarden/policy-server/blob/adea2081933a9bc301131d28849a62c396ea1255/src/settings.rs#L11-L19).

The `ClusterAdmissionPolicy` already has a `mutating` field. The `kubewarden-controller` when writing the `policy-server` configuration has to add this `mutating` field as well (has to be added in https://github.com/kubewarden/policy-server/blob/adea2081933a9bc301131d28849a62c396ea1255/src/settings.rs#L11-L19).

The `policy-server` knows for every instantiated policy whether it's expected to behave as mutating or not. No changes are required nor in any policy or in the SDK's.

There are different cases:

1. Mutating policy with `mutating: false` in the Custom Resource. This means that despite the policy has the ability to accept a request by mutating it, we don't want to mutate (a `ValidatingWebhookConfiguration` will be created). In this case, the `policy-server` has to reject the request with a message describing that the policy would have accepted the request by mutating it, but it will be rejected as requested.
2. Mutating policy with `mutating: true`. Nothing changes. A `MutatingWebhookConfiguration` will be registered, and it's up to the policy code to either accept, mutate or reject the request.
3. Non-mutating policy with `mutating: false`. Nothing changes. A `ValidatingWebhookConfiguration` will be registered, and it's up to the policy code to either accept or reject the request.
4. Non-mutating policy with `mutating: true`. Nothing changes. We can write an INFO message on stderr describing that the policy will never mutate, just for the user to know that they are expecting mutating from a non-mutating policy. In terms of functionality, the policy works as expected, despite it only validates.

jvanz:
I've proposed a fix for that. The related PR are:

- https://github.com/kubewarden/policy-server/pull/139
- https://github.com/kubewarden/policy-evaluator/pull/71
- https://github.com/kubewarden/kubewarden-controller/pull/141
- https://github.com/kubewarden/kubewarden-end-to-end-tests/pull/5

chrisns:
👋 happy new year etc :) any idea on when this can be merged in? 😄

jvanz:
> wave happy new year etc :) any idea on when this can be merged in? smile

Hi @chrisns ! Happy new year! I hope you've enjoyed good food and time with nice people during these days!

[...]

PR #139, opened 2021-12-31
 Adds mutating settings field.

jvanz:
Adds the "mutating" field in the settings. This allows policy server to check if a policy is allowed to mutate a request. If the policy mutate the request, but its "mutating" settings is false, the request is rejected.

Fixes: https://github.com/kubewarden/policy-server/issues/137.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 HYPERFINE := $(shell command -v hyperfine 2> /dev/null)
+IMG ?= policy-server:latest
 
 .PHONY: build
 build:
@@ -26,3 +27,7 @@ tag:
 	@git-chglog --output CHANGELOG.md
 	@git commit -m 'Update CHANGELOG.md' -- CHANGELOG.md
 	@git tag -f "${TAG}"
+
+.PHONY: docker-build
+docker-build: test ## Build docker image with the manager.
+	docker build -t ${IMG} .
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -10,7 +10,8 @@ use std::path::{Path, PathBuf};
 #[derive(Deserialize, Debug, Clone)]
 pub struct Policy {
     pub url: String,
-
+    #[serde(rename = "allowedToMutate")]
+    pub allowed_to_mutate: Option<bool>,
     #[serde(skip)]
     pub wasm_module_path: PathBuf,
 
@@ -65,8 +66,44 @@ example:
         assert!(!policies.is_empty());
 
         let policy = policies.get("example").unwrap();
+        assert!(policy.allowed_to_mutate.is_none());
         let settings = policy.settings();
         assert!(settings.is_some());
+    }
+
+    #[test]
+    fn test_allowed_to_mutate_settings() {
+        let input = r#"
+---
+example:
+  url: file:///tmp/namespace-validate-policy.wasm
+  allowedToMutate: true
+  settings:
+    valid_namespace: valid
+"#;
+        let policies: HashMap<String, Policy> = serde_yaml::from_str(input).unwrap();
+        assert!(!policies.is_empty());
+
+        let policy = policies.get("example").unwrap();
+        assert!(policy.allowed_to_mutate.unwrap());
+        let settings = policy.settings();
+        assert!(settings.is_some());
+
+        let input2 = r#"
+---
+example:
+  url: file:///tmp/namespace-validate-policy.wasm
+  allowedToMutate: false
+  settings:
+    valid_namespace: valid
+"#;
+        let policies2: HashMap<String, Policy> = serde_yaml::from_str(input2).unwrap();
+        assert!(!policies2.is_empty());
+
+        let policy2 = policies2.get("example").unwrap();
+        assert_eq!(policy2.allowed_to_mutate.unwrap(), false);
+        let settings2 = policy2.settings();
+        assert!(settings2.is_some());
     }
 
     #[test]
--- a/src/worker.rs
+++ b/src/worker.rs
@@ -5,7 +5,7 @@ use policy_evaluator::{
     policy_evaluator::{PolicyEvaluator, ValidateRequest},
     policy_evaluator_builder::PolicyEvaluatorBuilder,
     policy_metadata::Metadata,
-    validation_response::ValidationResponse,
+    validation_response::{ValidationResponse, ValidationResponseStatus},
 };
 use std::collections::HashMap;
 use std::fmt;
@@ -18,8 +18,13 @@ use crate::metrics;
 use crate::settings::Policy;
 use crate::utils::convert_yaml_map_to_json;
 
+struct PolicyEvaluatorWithSettings {
+    policy_evaluator: PolicyEvaluator,
+    allowed_to_mutate: Option<bool>,
+}
+
 pub(crate) struct Worker {
-    evaluators: HashMap<String, PolicyEvaluator>,
+    evaluators: HashMap<String, PolicyEvaluatorWithSettings>,
     channel_rx: Receiver<EvalRequest>,
 }
 
@@ -118,7 +123,12 @@ impl Worker {
                 continue;
             }
 
-            evs.insert(id.to_string(), policy_evaluator);
+            let policy_evaluator_with_settings = PolicyEvaluatorWithSettings {
+                policy_evaluator,
+                allowed_to_mutate: policy.allowed_to_mutate,
+            };
+
+            evs.insert(id.to_string(), policy_evaluator_with_settings);
         }
 
         if !evs_errors.is_empty() {
@@ -137,7 +147,10 @@ impl Worker {
             let _enter = span.enter();
 
             let res = match self.evaluators.get_mut(&req.policy_id) {
-                Some(policy_evaluator) => match serde_json::to_value(req.req.clone()) {
+                Some(PolicyEvaluatorWithSettings {
+                    policy_evaluator,
+                    allowed_to_mutate,
+                }) => match serde_json::to_value(req.req.clone()) {
                     Ok(json) => {
                         let start_time = Instant::now();
                         let resp = policy_evaluator.validate(ValidateRequest::new(json));
@@ -148,6 +161,21 @@ impl Worker {
                             status.code
                         } else {
                             None
+                        };
+                        let resp = if mutated && allowed_to_mutate.as_ref() == Some(&false) {
+                            ValidationResponse {
+                                allowed: false,
+                                status: Some(ValidationResponseStatus {
+                                    message: Some(format!("Request rejected by policy {}. The policy attempted to mutate the request, but it is currently configured to not allow mutations.", &req.policy_id)),
+                                    code: None,
+                                }),
+                                // if `allowed_to_mutate` is false, we are in a validating webhook. If we send a patch, k8s will fail because validating webhook do not expect this field
+                                patch: None,
+                                patch_type: None,
+                                ..resp
+                            }
+                        } else {
+                            resp
                         };
                         let res = req.resp_chan.send(Some(resp));
                         let policy_evaluation = metrics::PolicyEvaluation {


src/worker.rs
```diff
@@ -88,22 +89,23 @@ impl Worker {
...
-                Err(err) => {
-                    evs_errors.insert(
-                        policy.url.clone(),
-                        format!("could not instantiate policy: {:?}", err),
-                    );
-                    continue;
-                }
-            };
+            let mut policy_evaluator =
+                match PolicyEvaluatorBuilder::new(id.to_string(), policy.mutating)
```
jvanz, line 93: I'm not very happy with the necessity to send the flag for the `PolicyEvaluatorBuilder`. But I could not find a better solution for that. Considering the time that I have before vacations (today is 12/31/2021) I'm not sure if I find be able to find a better solution in time. So, I'm opening the current code to feedback from other colleagues.

Makefile
```diff
@@ -26,3 +27,12 @@ tag:
...
 	@git commit -m 'Update CHANGELOG.md' -- CHANGELOG.md
 	@git tag -f "${TAG}"
+
+.PHONY: docker-build
+docker-build: test ## Build docker image with the manager.
+	docker build -t ${IMG} .
+
+.PHONY: docker-push
+docker-push: ## Push docker image with the manager.
+	docker push ${IMG}
```
flavio, line 37: nitpicking, do we really need these tasks? I'm concerned someone might accidentally push the `policy-server:latest` to GH from his local machine...

src/settings.rs
```diff
@@ -10,6 +10,7 @@ use std::path::{Path, PathBuf};
 #[derive(Deserialize, Debug, Clone)]
 pub struct Policy {
     pub url: String,
+    pub mutating: bool,
```
flavio, line 13: I think we should name it in a different way. The policy server doesn't need to know if every policy is mutating or not. It can be made aware of **certain** policies that have to be "downgraded" from mutating to validating.

I would change this attribute to be named something like `downgrade_to_validation`. This would be `false` by default, which is good -> not specifying a value for that means "let the policy do what the author intended".

src/worker.rs
flavio: I think this is fine, I would however make that a dedicated method of the policy (`downgrade_to_validation()` with no params).

As written above, I think we should by default do what the policy author wanted, and then in special cases force this downgrade. Basically, the settings entries should by default fall back to do what the policy author intended.

You will see that as part of the GH review I'm authoring

src/worker.rs
```diff
@@ -149,7 +151,19 @@ impl Worker {
                         } else {
                             None
                         };
-                        let res = req.resp_chan.send(Some(resp));
+                        let res;
+                        if mutated && !policy_evaluator.policy.mutating {
+                            let error_msg = "Non mutating policy tried to mutate a request. The request has been rejected.".to_string();
```
flavio, line 156: This is the message that is shown to the final user. We should report something better/less scary.

> Request rejected by policy <policy name>. The policy attempted to mutate the request, but it is currently configured to not allow mutations.

Moreover, I would also emit a trace with what the mutation would look like. This can be useful for debugging purposes.

flavio (changes requested):
Overall ok. As expressed inside of the review, I think we should not require the settings file to be explicit about the behaviour of each policy (it is mutating or not?).

By default nothing is specified -> the policy outcome validate/reject/mutate is just sent back to the API server.

The only, optional, value added to the settings is whether the mutation results should be downgraded to rejections.

src/worker.rs
flavio: Maybe a better name could be `disable_mutations`...

src/settings.rs
ereslibre: I would be more explicit if we are not going with `mutate: bool`. I would say `allowed_to_mutate: bool` instead.

flavio: naming is always tough :)

I like `allowed_to_mutate` too

Makefile
viccuad: Agree here. It's reassuring to use `policy-server:latest` and know that it comes from main and tests passed. Maybe drop the `docker-push` target? The `docker-build` is useful for local dev.

src/worker.rs
viccuad: +1!

src/settings.rs
```diff
@@ -46,6 +47,7 @@ mod tests {
 ---
 example:
   url: file:///tmp/namespace-validate-policy.wasm
+  mutating: false
```
viccuad, line 50: If we move to an optional configuration, that defaults to false, then I would allow for the configuration entry to be missing, for backwards compatibility. I would then remove all the `mutating: false` on the tests, and add just a new test with it set to true.

src/worker.rs
ereslibre: I think we can avoid adding this field to the PolicyEvaluatorBuilder. I think it has to be abstracted from the evaluator itself, given it's not its responsibility. This is my proposal on top of this PR: https://gist.github.com/ereslibre/819d94af9445a28ca761fb1d1111871c

ereslibre (changes requested):
I think this setting is not strictly tied to the PolicyEvaluator but how we treat an answer from it based on higher level settings (`allowed_to_mutate`).

I attach a proposal on how to tackle this problem.

src/worker.rs
ereslibre: Also, I agree that the setting can be optional and default to the policy behavior if it's omitted.

flavio:
This PR fixes https://github.com/kubewarden/allow-privilege-escalation-psp-policy/issues/9

Makefile
jvanz: I used these targets to build and publish the container image later used my tests scripts. But I understand the concerns. Let me change this.

src/worker.rs
jvanz: I've updated the branch with the changes suggested by @ereslibre

src/settings.rs
```diff
@@ -10,7 +10,8 @@ use std::path::{Path, PathBuf};
 #[derive(Deserialize, Debug, Clone)]
 pub struct Policy {
     pub url: String,
-
+    #[serde(default)] // default bool value is "false"
+    pub allowed_to_mutate: bool,
```
ereslibre, line 14: I think the plan was to have something like:

```suggestion
    #[serde(default)]
    pub allowed_to_mutate: Option<bool>,
```

If `None` is provided (it wasn't specified in the config file explicitly), then default `allowed_to_mutate` to the policy metadata mutating field.

src/settings.rs
```diff
@@ -10,7 +10,8 @@ use std::path::{Path, PathBuf};
 #[derive(Deserialize, Debug, Clone)]
 pub struct Policy {
     pub url: String,
-
+    #[serde(default)] // default bool value is "false"
```
ereslibre, line 13: Here the default value will be None, we can remove this comment.

Makefile
```diff
@@ -1,4 +1,5 @@
 HYPERFINE := $(shell command -v hyperfine 2> /dev/null)
+IMG ?= policy-server:local
```
ereslibre, line 2: I don't think we need the `docker-build` target, but if we add it, I would change the default tag here. It is not coming from a repo such as ghcr.io already.

```suggestion
IMG ?= policy-server:latest
```

src/worker.rs
```diff
@@ -149,6 +162,21 @@ impl Worker {
...
+                                allowed: false,
+                                status: Some(ValidationResponseStatus {
+                                    message: Some(format!("Request rejected by policy {}. The policy attempted to mutate the request, but it is currently configured to not allow mutations.", &req.policy_id)),
+                                    code: None,
+                                }),
+                                ..resp
+                            }
+                        } else {
+                            resp
+                        };
```
ereslibre, line 179: I think what follows is a bit more readable:

```suggestion
                        let resp = if mutated && allowed_to_mutate == Some(false)
                        {
                            ValidationResponse {
                                allowed: false,
                                status: Some(ValidationResponseStatus {
                                    message: Some(format!("Request rejected by policy {}. The policy attempted to mutate the request, but it is currently configured to not allow mutations.", &req.policy_id)),
                                    code: None,
                                }),
                                ..resp
                            }
                        } else {
                            resp
                        };
```

ereslibre (changes requested):
LGTM besides two small comments. Thank you!

src/settings.rs
```diff
@@ -65,10 +66,46 @@ example:
         assert!(!policies.is_empty());
 
         let policy = policies.get("example").unwrap();
+        assert_eq!(policy.allowed_to_mutate, Option::default());
```
flavio, line 69: nitpicking, maybe inside of the test it would make more sense to be explicit about what we expect the value to be:
replace  `Option::default()` with `None`.

src/worker.rs
```diff
@@ -18,8 +18,13 @@ use crate::metrics;
 use crate::settings::Policy;
 use crate::utils::convert_yaml_map_to_json;
 
+struct PolicyEvaluatorWithSettings {
+    policy_evaluator: PolicyEvaluator,
+    allowed_to_mutate: Option<bool>,
+}
```
flavio, line 24: I would change the `PolicyEvaluator` struct to have the optional `allowed_to_mutate` field. In this way, if we will ever want to offer this function also to kwctl, we will have all the things we need.

Plus, I personally prefer to have this attribute into `PolicyEvaluator`, rather then create a wrapper to hold it.

flavio (changes requested):
Overall ok, I left some comments

src/worker.rs
ereslibre: I thought we agreed on not going in that direction (https://github.com/kubewarden/policy-evaluator/pull/71).

If you folks agree that this should go in the policy-evaluator go ahead. I don't think it's the case, given I believe this is a behavior of what to do with the result of the policy-evaluator.

flavio: Yeah, scratch it. I'm fine with the current implementation of this PR

src/worker.rs
```diff
@@ -149,6 +166,20 @@ impl Worker {
                         } else {
                             None
                         };
+                        let resp = if mutated && disallowed_mutation(allowed_to_mutate) {
```
jvanz, line 169: @ereslibre , your suggestion to make this `if` more readable does not compile due a issue of comparing different types (`&mut Option<bool>` and `Option<bool>`). One solution is to implement the `PartialEq` trait for the types. But I was not sure if implementing the trait for type from std lib is a best practice or not. What do you think? 

Anyway, trying to improve the readability I've extracted the code into a function.

src/worker.rs
jvanz: @ereslibre , your suggestion to make this if more readable does not compile due a issue of comparing different types (&mut Option<bool> and Option<bool>). One solution is to implement the PartialEq trait for the types. But I was not sure if implementing the trait for type from std lib is a best practice or not. What do you think?

Anyway, trying to improve the readability I've extracted the code into a function.

src/worker.rs
ereslibre: You can do:

```rust
let resp = if mutated && allowed_to_mutate.as_ref() == Some(&false) {
```

I think is a bit more clear than the utility function.

jvanz: :facepalm: I need to put more rust in daily routine...

jvanz: Done!

ereslibre (approved):
LGTM besides the comment. Feel free to merge from my side.

src/worker.rs
```diff
@@ -149,6 +162,21 @@ impl Worker {
...
                             None
                         };
+                        let resp = if mutated && allowed_to_mutate.as_ref() == Some(&false) {
+                            ValidationResponse {
+                                allowed: false,
+                                status: Some(ValidationResponseStatus {
+                                    message: Some(format!("Request rejected by policy {}. The policy attempted to mutate the request, but it is currently configured to not allow mutations.", &req.policy_id)),
+                                    code: None,
+                                }),
+                                // if the mutate is false, we are in a validating webhook. If we send a patch, k8s will failed. Because validating webhook should not has patch
```
flavio, line 172: grammar fix:

```suggestion
                                // if `allowed_to_mutate` is false, we are in a validating webhook. If we send a patch, k8s will fail because validating webhook do not expect this field
```

flavio (commented):
Just fix the sentence I pointed out and merge it :rocket:

flavio (approved):

Issue #98, opened 2021-10-05
Allow verification of policies using sigstore


flavio:
Kubewarden policies are WebAssembly modules stored inside of OCI registries. Users can reference them by shasum if wanted, but it would way better to provide a stronger way to verify the policies before using them.

[sigstore](https://sigstore.dev) is a new standard that allows the signature and verification of container images and OCI artifacts.

We should leverage sigstore to provide a way to verify policies before their execution.

## Use case

> As an administrator,
> I want all the Kubewarden policies loaded in production to be signed with a certain key,
> so that unverified policies are not loaded.

## How that should work

* Policies are signed using `cosign` or some equivalent tool. 
* Policy Server has a configuration flag that enforces all the policies mentioned inside of its configuration file to be signed
* PolicyServer reads a list of keys and annotations from a verification.yaml file. This file will be mounted as a Secret for PolicyServer to use it. Example:
```yaml
verification-keys:
  key-name-irrelevant: |
        -----BEGIN PUBLIC KEY-----
        MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEX0HFTtCfTtPmkx5p1RbtwDE1EVzu
        wjQs1cCRKb5Pz/yUspkQsN3FO4iyWodCy5j3o0CdIJD/1gvq98pf4IG9tA==
        -----END PUBLIC KEY-----
verification-annotations: # optional
  env: prod
```
* Policy Server exits with an error if one or more policies fail the verification

Note well: this is a rough idea about how this feature should work. We definitely need to spend more time discussing how that should work. Feedback is welcome!

viccuad:
Given that each policy can be signed with a different key and a list of annotations, I will go with the approach of adding those fields as optional to policy-server's policies.yaml:
```yaml
psp-apparmor:
  url: registry://ghcr.io/kubewarden/policies/psp-apparmor:v0.1.3
psp-capabilities:
  url: registry://ghcr.io/kubewarden/policies/psp-capabilities:v0.1.2
  verification-keys:
    - key: |
         pub key
  verification-annotations:
      - env: prod
  settings:
    allowed_capabilities: ["*"]
    required_drop_capabilities: ["KILL"]
```
I expect these fields to be present in the ClusterAdmissionPolicy CRD in the future.

[...]

PR #140, opened 2022-01-11
Bring back policy verification using Sigstore

flavio:
This PR brings back the code that was previously reverted from the main branch.

The code now builds with the latest version of our internal crates (like policy-fetcher, where all the verification code lies).

Note well: currently both Fulcio's certificate and Rekor's public key are hard coded into the sources of Policy Server.
This is a temporary workaround until sigstore-rs allows to download this data from Sigstore's TUF repository.

Finally, the UX of policy verification has been slightly improved.
The verification of policies is automatically turned on when the user provides the `--verification-path` flag.
Prior to this commit, the user had to use both the `--enable-verification` and the `--verification-path` flags.


src/main.rs
```diff
@@ -186,6 +202,69 @@ fn main() -> Result<()> {
...
             debug!(policy = name.as_str(), "download");
+
+            let mut verifier = match policy_fetcher::verify::Verifier::new(
+                sources.clone(),
+                FULCIO_CRT.as_bytes(),
+                REKOR_PUB_KEY,
+            ) {
+                Ok(v) => v,
+                Err(e) => {
+                    fatal_error(format!("Cannot created sigstore verifier: {:?}", e));
```
viccuad, line 213: Little typo:

```suggestion
                    fatal_error(format!("Cannot create sigstore verifier: {:?}", e));
```

viccuad (approved):
LGTM, thanks. Approving, just a little typo.

viccuad:
Fixes https://github.com/kubewarden/policy-server/issues/98

PR #141, opened 2022-01-12
fix: Don't default to verification always enabled

viccuad:
`matches.is_present("verification-path")` is always true, as the flag has `.default("verification.yml")`. Remove default.

Tested locally with `cargo run --release -- --policies policies.yml --workers 2 --log-fmt otlp --log-level debug`, and the `--verification-path` flag.

Fixes https://github.com/kubewarden/policy-server/issues/98

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -113,7 +113,6 @@ pub(crate) fn build_cli() -> App<'static, 'static> {
             Arg::with_name("verification-path")
                 .env("KUBEWARDEN_VERIFICATION_CONFIG_PATH")
                 .long("verification-path")
-                .default_value("verification.yml")
                 .help("YAML file holding verification information (URIs, keys, annotations...)"),
         )
         .arg(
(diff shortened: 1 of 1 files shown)


viccuad:
This looks shoddy, but it tries to be a fix, instead of feat. The behavior is:
- Call `policy-server` with no verification flags: no verification performed. Default `verification.yml` is not honoured not used.
- Call `KUBEWARDEN_ENABLE_VERIFICATION=true policy-server`: Verification performed, `verification.yml` is used.
- Call `KUBEWARDEN_ENABLE_VERIFICATION=true policy-server --verification-path=foo.yml`: Verification performed, `foo.yml` is used.
- Call `policy-server --verification-path=foo.yml`: Verification performed, `foo.yml` is used.

flavio:
Ouch, thanks for the fix. I definitely missed that.

I'm not convinced about the fix however. The use case I had in mind was the following one:

```console
policy-server --verification-path=foo.yml or policy-server --enable-verification --verification-path=verification.yml
```

What if we remove the default value of `verification-path` from `clap`'s definition? In that way we could cover the following scenarios:

**1. Verification using keys:**

```console
policy-server --verification-path=foo.yml or policy-server --enable-verification --verification-path=verification.yml
```

Note well: we save some typing by not requiring `--enable-verification`

**2. Keyless verification**

This is not yet implemented, but it's going to be our "next thing":

```console
policy-server --enable-verification--keyless or policy-server --enable-verification --verification-path=verification.yml
```

I thought a bit about what to do in that case. I think this is worth to have a dedicated flag because maybe the end user doesn't want to force the verification of signatures produced in keyless mode.

**3. Key based verification + Keyless verification**

Nothing surprising there:

```console
policy-server --enable-verification-keyless or policy-server --enable-verification --verification-path=verification.yml
```

If we want to go ahead with this approach we can also remove the `--verification--enable` flag we currently have.

What do you think?

viccuad:
Force-pushed, dropping the default value for `--verification-path`, and allowing for `policy-server --verification-path=foo.yml` to perform validation. Updated PR description.

As discussed, we can in the future add keyless info such as openIDC ids to the config file passed in `--verification-path`, which would perform keyless verification.
This works nicely with env vars, and users of Kubewarden K8s stack can activate it via the Configmap containing the verification config.

flavio (approved):
LGTM, thanks for having gone through all the design iteration :)

PR #175, opened 2022-02-01
fix: Enable verification if `--verification-path` is passed

viccuad:
This was dropped in the clap bump.

With this, the correct behaviour is back:
- use verification if `--enable-verification` or `KUBEWARDEN_ENABLE_VERIFICATION=1` is present.
-  use verification if `--verification-path=foo` is present (expected by controller).


ereslibre (approved):
Ouch, LGTM.

PR #193, opened 2022-02-14
Do not flatten settings

ereslibre:
Do not flatten the `settings` field that would behave as a "catch-all"
for unknown attributes.

This would silently make the `policy-server` forward certain
attributes to policies when keys change on the configuration. We
should plan for such changes and have a migration path.

Policies should only receive as `settings` what is intended as such.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 
 use serde::Deserialize;
-use serde_yaml::{Mapping, Value};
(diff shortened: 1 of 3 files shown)


flavio:
I think that makes sense. This means we will have to change how kubewarden-controller produces the `policies.yml` file

jvanz (approved):
That's cool, thanks!

flavio (approved):
LGTM, nice cleanup!

Issue #190, opened 2022-02-10
Implement `monitor` mode


ereslibre:
RFC: https://github.com/kubewarden/kubewarden-controller/blob/main/rfc/0002-monitor-mode.md

Implement the `monitor` mode in the `policy-server`. This involves several changes:

1. Extend the [`Policy` struct](https://github.com/kubewarden/policy-server/blob/e0d97f431b528bb268869e010a547f40f4804d31/src/settings.rs#L10-L20) used for settings so that it includes a new optional `mode` attribute. Valid values are `monitor` and `protect`. It defaults to `protect` if empty.

2. Add `mode` to the `PolicyEvaluation` metric baggage.

3. Whenever an evaluation is requested, if the policy has a `monitor` `mode`, log the request with `INFO` level. If the policy is `mutating` and is configured as `mutating`, behave in the following way:
    2.1. If the request was `accepted` or `rejected`, `accept` the request, logging the event (counted in the policy evaluation metrics as normal).
    2.2. If the request was to accept with mutation, `accept` the request -- without mutating --, logging the event [and the patch that would have been returned to the API server].

**Note**: is important to understand that this is not a perfect behavior, because mutations are not happening -- just logged --. This means, that validating policies (that come after mutating ones), might be evaluating resources that are different to what they would be in production.

PR #194, opened 2022-02-14
Monitor mode

ereslibre:
Fixes: https://github.com/kubewarden/policy-server/issues/190

feat: implement policy modes

Policies can be configured to either be in `monitor` or `protect`
modes. When a policy is in `protect` mode, it will accept, reject or
mutate requests. When a policy is in `monitor` mode, it will always
accept all requests, but will log what would have been its decision if
it had been in `protect` mode.

This is helpful if you want to check how a policy would behave before
it starts to take decisions on a Kubernetes cluster.

### DIFF
--- a/src/worker.rs
+++ b/src/worker.rs
@@ -11,16 +11,17 @@ use std::collections::HashMap;
 use std::fmt;
 use std::time::Instant;
 use tokio::sync::mpsc::{Receiver, Sender};
-use tracing::{error, info_span};
+use tracing::{error, info, info_span};
 
 use crate::communication::EvalRequest;
 use crate::metrics;
-use crate::settings::Policy;
+use crate::settings::{Policy, PolicyMode};
 use crate::utils::convert_yaml_map_to_json;
 
 struct PolicyEvaluatorWithSettings {
     policy_evaluator: PolicyEvaluator,
-    allowed_to_mutate: Option<bool>,
+    policy_mode: PolicyMode,
+    allowed_to_mutate: bool,
 }
 
 pub(crate) struct Worker {
@@ -125,7 +126,8 @@ impl Worker {
 
             let policy_evaluator_with_settings = PolicyEvaluatorWithSettings {
                 policy_evaluator,
-                allowed_to_mutate: policy.allowed_to_mutate,
+                policy_mode: policy.policy_mode.clone(),
+                allowed_to_mutate: policy.allowed_to_mutate.unwrap_or(false),
             };
 
             evs.insert(id.to_string(), policy_evaluator_with_settings);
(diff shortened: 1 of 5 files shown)


jvanz (approved):
LGTM! :D

src/worker.rs
```diff
@@ -141,6 +143,62 @@ impl Worker {
...
+    //   configuration does not allow it to mutate
+    // - A policy might be running in "Monitor" mode, that always
+    //   accepts the request (without mutation), logging the answer
+    fn validation_response_with_constraints(
+        policy_id: &str,
+        policy_mode: &PolicyMode,
+        allowed_to_mutate: bool,
+        validation_response: ValidationResponse,
+    ) -> ValidationResponse {
+        if validation_response.patch.is_some() && !allowed_to_mutate {
```
flavio, line 158: Shouldn't we look at `policy_mode` earlier inside of this method?

Let's assume I declare a policy with the following attributes:
  * allowed to mutate: false
  * policy mode: monitor

Let's assume the evaluation of this policy would lead to a mutation.

This would lead Policy Server to reply with a rejection message. What I expect is to have the policy accept the object, with no mutation done (because `allowed_to_mutate` is `false`)

src/worker.rs
```diff
@@ -149,37 +207,34 @@ impl Worker {
...
+                            &req.policy_id,
+                            &policy_mode,
+                            allowed_to_mutate,
+                            validation_response,
+                        );
+                        let res = req.resp_chan.send(Some(validation_response));
                         let policy_evaluation = metrics::PolicyEvaluation {
-                            policy_name: policy_evaluator.policy.id.clone(),
+                            policy_name,
+                            policy_mode: policy_mode.into(),
```
flavio, line 237: I think we have to do more changes. Right now the metrics are unable to distinguish between protecting policies accepting requests, monitoring policies that would one day accept requests and - most important of all - monitoring policies that would one day reject a request.

I think that when dealing with monitoring policies, we should retain the "future status when protect mode is enabled", and expose that inside of the metrics.
To simplify, we should emit a metric event about a rejection. The reason, IMHO, is to allow an operator to quickly identify the monitoring policies that, once switched to protect mode, would instead generate a flood of rejections.

Right now, this information is buried only inside of jaeger traces

flavio (commented):
Overall good, I left some questions/comments

src/worker.rs
ereslibre: Being this part of the baggage of the metric, the user can use this `policy_mode` to filter the evaluation metric done in different modes, and what they would have answered in the case of `policy_mode: monitor`.

src/worker.rs
flavio: I don't think this would be a misconfiguration. The way I see it, a user who takes a mutating policy and marks that as validating only is doing that on purpose.
That's why I say we should still accept the request

src/worker.rs
flavio: The metric is exposing the value of `accepted`, which is set on line 226:

```rust
let accepted = validation_response.allowed;
```

The `validation_response` object is returned by the function `validation_response_with_constraints`.

Inside of `validation_response_with_constraints`, a policy that has `monitor` mode will always return a response with `allowed` set to `true`. The information about the possible outcome when the policy is switched to "protect" mode is lost.

src/worker.rs
ereslibre: Let me circle back a bit then. What you specify here:

"What I expect is to have the policy accept the object, with no mutation done (because allowed_to_mutate is false)"

is not what is going to happen. In the case you have described, the policy server will reject the request with an error message, saying the policy tried to mutate a request while it was configured as validating only.

flavio: Oh yes, that makes sense! Sorry about the noise and thanks for the explanation

src/worker.rs
ereslibre: Hm, no, it's not lost. Line 226 goes all the way to 242 (the metric baggage)... Am I missing something?

ereslibre: Just to give more information as well: `validation_response_with_constraints` does not mutate its input. It builds a new `ValidationResponse` object, that copies some arguments inside its structure attributes, but is a new object.

flavio: No, it's my fault. Looking at the whole file I see what is happening. I got mislead by the github diff view + the overwriting of the `validation_response`.

It's all good... sorry about the noise

flavio (approved):
LGTM

src/worker.rs
ereslibre: Thanks 🙏

PR #207, opened 2022-03-03
feat: add info to traces about why a resource was rejected by a policy

### LINKED ISSUE #206: Feature Request: add info to traces on why a resource was rejected by a policy

There is no information about why a resource was rejected by a policy in the traces.

### Solution you'd like

Display validation error message and code in the traces, then the reason of the rejection is displayed when seeing the traces inside a tool like Jaeger.

_No response_

### Anything else?

_No response_

raulcabello:
## Description

add info to traces about why a resource was rejected by a policy

Fix #206

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -57,6 +57,14 @@ fn populate_span_with_admission_request_data(adm_req: &AdmissionRequest) {
 fn populate_span_with_policy_evaluation_results(validation: &ValidationResponse) {
     Span::current().record("allowed", &validation.allowed);
     Span::current().record("mutated", &validation.patch.is_some());
+    if let Some(status) = &validation.status {
+        if let Some(code) = &status.code {
+            Span::current().record("response_code", code);
+        }
+        if let Some(message) = &status.message {
+            Span::current().record("response_message", &message.as_str());
+        }
+    }
 }
 // note about tracing: we are manually adding the `policy_id` field
 // because otherwise the automatic "export" would cause the string to be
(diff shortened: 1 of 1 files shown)


src/api.rs
```diff
@@ -57,6 +57,13 @@ fn populate_span_with_admission_request_data(adm_req: &AdmissionRequest) {
 fn populate_span_with_policy_evaluation_results(validation: &ValidationResponse) {
     Span::current().record("allowed", &validation.allowed);
     Span::current().record("mutated", &validation.patch.is_some());
+    if let Some(status) = &validation.status {
+        Span::current().record("response_code", &status.code.unwrap_or_default());
+        Span::current().record(
+            "response_message",
+            &status.message.clone().unwrap_or_default().as_str(),
+        );
```
ereslibre, line 65: I think is better to `record` only if they have a `Some`, something like (pseudorust):

```rust
if let Some(response_code) = status.code {
  Span::current().record("response_code", response_code);
}
if let Some(response_message) = status.message {
  Span::current().record("response_code", response_message);
}
```

This way, if they are not filled, they are kept as `tracing::field::Empty` as opposed to an empty string. This ensures that it's not recorded altogether, as opposed to being recorded as an empty string.

ereslibre (changes requested):
Thanks @raulcabello!, a comment regarding what to do with missing attributes

src/api.rs
raulcabello: good point! done in https://github.com/kubewarden/policy-server/commit/04ace1da7eca104f51d0e81634fa8b2a734bf1d1

ereslibre (approved):
Thanks @raulcabello! 👏

jvanz (approved):
LGTM! Thanks! :D

Issue #211, opened 2022-03-11
Feature Request: improve keyless verification by fetching Fulcio and Rekor data from remote TUF repository


flavio:
### Is your feature request related to a problem?

This issue is about having Policy Server fetch Fulcio's certificates and Rekor's public key from the official TUF repository used by the sigstore project.

### Solution you'd like

Right now we have these files bundled inside of our source code, but we should move away from this approach.

The `sigstore-rs` crate offers a way to obtain these files by interacting with sigstore's official TUF repository.

Given policy-server is seldomly restarted, we will not cache these files to disk during this first iteration.
Moreover, during this first iteration we will not add flags that allow the user to provide custom files (like kwctl does).

PR #214, opened 2022-03-14
feat: Add keyless verification by supporting verification-config.yml

raulcabello:
Implement Sigstore keyless verification for policy server.
Now, accept new --verification-path that contains the config from [the Sigstore verification RFC](https://github.com/kubewarden/kubewarden-controller/blob/main/rfc/0003-policy-signing.md).

Fix #210

PR #215, opened 2022-03-15
improve policy verification

flavio:
> This PR supersedes https://github.com/kubewarden/policy-server/pull/214

- feat: Add keyless verification by supporting verification-config.yml
- feat: fetch sigstore data from remote TUF repo

Fixes https://github.com/kubewarden/policy-server/issues/211


src/main.rs
```diff
@@ -179,60 +171,49 @@ fn main() -> Result<()> {
...
+                        unreachable!()
+                    }
+                };
+                verified_manifest_digest = Some(
+                    ver.verify(&policy.url, docker_config.clone(), verification_config)
+                        .await
+                        .map_err(|e| {
+                            fatal_error(format!("Policy '{}' cannot be verified: {:?}", name, e))
+                        })
+                        .unwrap(),
```
ereslibre, line 215: In case of more than one offender this is only reporting the first one and exiting, right? I think it would be nice to report all offenders at once.

raulcabello (commented):
Are we going to allow users to use their own Fulcio certificate as we do in kwctl? I see we always fetch the code from TUF `FulcioAndRekorData::FromTufRepository` , but there is no way to use your own certificate `FulcioAndRekorData::FromCustomData`

src/main.rs
viccuad: No, it prints all the constraints from the verification_config that are not satisfied. E.g:

https://github.com/kubewarden/policy-fetcher/blob/6fe85ba4f7a64e60171aead5511cea34d78d261d/src/verify/mod.rs#L526-L539

src/cli.rs
```diff
@@ -183,28 +184,16 @@ pub(crate) fn policies(matches: &clap::ArgMatches) -> Result<HashMap<String, Pol
     })
 }
 
-pub(crate) fn verification_settings(matches: &clap::ArgMatches) -> Result<VerificationSettings> {
-    if matches.value_of("verification-path").is_none() {
-        return Err(anyhow!(
+pub(crate) fn verification_settings(
```
viccuad, line 187: Nit:

```suggestion
pub(crate) fn verification_config(
```
to match policy-fecther (and later variable)

viccuad (approved):
LGTM, besides small nit

src/main.rs
raulcabello: yes, but it will exit (fatal_error) as soon as one policy fails without evaluating the others policies pending in the loop

flavio:
I've updated the PR to reflect the changes you requested. I've converted the PR to be a draft because the policy-fetcher and policy-evalutator dependencies need to be updated. This is currently blocked by https://github.com/kubewarden/policy-fetcher/pull/69

flavio:
> Are we going to allow users to use their own Fulcio certificate as we do in kwctl? I see we always fetch the code from TUF `FulcioAndRekorData::FromTufRepository` , but there is no way to use your own certificate `FulcioAndRekorData::FromCustomData`

Yes, this isn't supported yet. I would add that via another PR.  This is something I don't see as super important right now for Policy Server.

Moreover, there's something that needs to be sorted out: we must provide a way to specify multiple fulcio certs via the cli flag. However, I have no idea about how this could be done via environment flags. Right now our controller uses environment variables as a way to configure the behaviour of policy-server. I don't know how we could expose that...

flavio:
This can be reviewed now. I'm working on adding some unit tests for the downloader code

flavio:
The unit tests are ready

flavio:
This is now building fine, because all the dependencies are now available :partying_face:

PR #235, opened 2022-04-21
gh test action: authenticate against GHCR

flavio:
GHCR must have some abuse system in place that blocks lots of connections done in an anonymous way. This seems to be causing our GH action testing policy-server to fail.

This commit ensures we perform authenticate operations against GHCR. The GITHUB_TOKEN is used to authenticate against ghcr.io

The testing done on my fork seems to prove this change helps. Unfortunately we will not know for sure until we merge this fix here.

### DIFF
--- a/src/policy_downloader.rs
+++ b/src/policy_downloader.rs
@@ -235,7 +235,9 @@ async fn create_verifier(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use directories::BaseDirs;
     use lazy_static::lazy_static;
+    use policy_fetcher::registry::config::read_docker_config_json_file;
     use std::sync::Mutex;
     use tempfile::TempDir;
     use tokio::runtime::Runtime;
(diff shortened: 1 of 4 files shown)


src/policy_downloader.rs
```diff
@@ -245,7 +247,20 @@ mod tests {
...
+            let docker_config = if let Some(dcp) = docker_config_path {
+                if dcp.exists() {
+                    Some(read_docker_config_json_file(&dcp).expect("Error reading docker config file"))
+                } else {
+                    None
+                }
+            } else {
+                None
+            };
+            //let docker_config = read_docker_config_json_file(path: &Path)
```
flavio, line 262: I just noticed this leftover, I'll remove it once the 1st run of unit test completes. I just want to see the results :crossed_fingers:

flavio:
Yes, all the tests are green :partying_face:

I'll push some esthetic fixes, let's hope everything will be green again :crossed_fingers:

flavio:
Nope, this is now broken...

flavio:
I think we should rewrite the test to use a mock, like @raulcabello and @viccuad just did with their sigstore PRs

Another approach, would be to run a docker registry next to the tested code, by using [testcontainers](https://docs.rs/testcontainers/0.13.0/testcontainers/). This is what oci-distribution uses to implement reliable e2e tests

viccuad:
Argh, what a find. Thanks. If we are going to do mocks, it would be nice to get them into testcontainers, so we can reuse them easier in other places. I suppose they would be accepted upstream, but I think testcontainers doesn't implement the concept of layers yet..

raulcabello:
For unit tests I would mock all network calls we do in `download_policies`, these test would run very fast and are more reliable.

For integration or e2e tests we can use testcontainers

flavio:
Let's go with the mock approach, as discussed during the daily :+1:

flavio:
I don't see this issue happening anymore. Closing the PR, we can reopen it later on if needed

Issue #99, opened 2021-10-05
Allow policy authors to write policies that perform image verification using sigstore


flavio:
[sigstore](https://sigstore.dev) is a new standard that allows the signature and verification of container images and OCI artifacts.

It would be great to allow policy authors to write policies that perform verification of container images using sigstore.

There's however one caveat: the sigstore code requires some data to be fetched from an OCI registry. Currently WASM code cannot make network calls.

To solve this problem, kubewarden's [policy-evaluator](https://github.com/kubewarden/policy-evaluator/) should expose a waPC function that the WASM guest can invoke. This function would run on the host, and it would take care of making the network requests and performing the verification.

PR #238, opened 2022-04-27
feat: Allow policy authors to write policies that perform image verification using sigstore

raulcabello:
## Description

pass fulcio_and_rekor_data to the Callback handler builder, since it is now needed for running policies that uses sigstore verification

Fix https://github.com/kubewarden/policy-server/issues/99


flavio (approved):
LGTM, of course we have to use the right policy-evaluator repository, not your fork

Issue #236, opened 2022-04-22
Feature Request:  Allow DNS Lookups from kubewarden Modules


floriankoch:
### Is your feature request related to a problem?

I try to validate if all Hosts inside a Ingress have a corresponding DNS Entry

### Solution you'd like

I like to use e.g. the Golang Function net.LookupIP to resolve the Domains

flavio:
This is the list of changes that have to be done:

* [ ] extend policy evaluator: https://github.com/kubewarden/policy-evaluator/pull/119
* [ ] extend rust SDK: https://github.com/kubewarden/policy-sdk-rust/pull/36
* [ ] extend kwcl: https://github.com/kubewarden/kwctl/pull/212
* [ ] extend policy-server
* [ ] extend go SDK
* [ ] extend Swift SDK

floriankoch:
@flavio   great news thank you

PR #243, opened 2022-04-29
support wapc dns lookup

flavio:
Support policies performing revers DNS lookups

Update to latest version of policy-evaluator, so that new waPC functions are supported.

This is needed to fix https://github.com/kubewarden/policy-server/issues/236

Note well: this builds on top of https://github.com/kubewarden/policy-server/pull/242


viccuad (approved):
(waiting for cargo.toml change)

flavio:
I've updated the `Cargo.toml` to reference the just released version of `policy-evaluator`. I'm going to merge it once all the tests are green

Issue #246, opened 2022-05-04
Add policy server namespace optional argument


ereslibre:
Add policy server namespace (`--policy-server-namespace` CLI arg and `POLICY_SERVER_NAMESPACE` envvar) that specifies on what namespace this policy server is running on.

This argument is optional. If it's not provided, the mechanism that follows will **not** be followed. So that the behavior of the `policy-server` is not unchanged compared to now if this flag or envvar is not provided.

So, if this namespace is provided:

- If the request is on a cluster-wide resource, continue normally.
- If the request is on a namespaced resource,
  - If the namespace does not match the provided namespace, continue normally.
  - If the namespace does match the provided namespace, accept the request.

This allows the Kubewarden stack to be permissive with a certain namespace (e.g. the namespace where all policy servers are created), and to also be strict if the user desires to do so. In the first model of operation, very widely applied policies will not impact the operation of Kubewarden itself. If the user/consumer is very expert and wants to prioritize security even more, they can still be strict and not provide this argument, what will not alter the `policy-server` behavior.

flavio:
I would prefer a different name to be used for this flag/env variable, something that makes more evident the behaviour that is going to be enabled.

What about having `--ignore-events-from-namespace` and `POLICY_SERVER_IGNORE_EVENTS_FROM_NAMESPACE`?

PR #247, opened 2022-05-06
Add option to always accept admission reviews on a given namespace

ereslibre:
Fixes: https://github.com/kubewarden/policy-server/issues/246

Add an `--always-accept-admission-reviews-on-namespace` option that
will always allow requests on the given namespace. Mutations are not
altered, and are allowed to happen in this namespace. However, all
evaluations of all policies will be evaluated to `allowed: true` if
provided.

This argument is optional, and by default, if not provided, the
behavior of the `policy-server` is unchanged.

This argument can also be provided as an environment variable named
`ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE`.

This argument only accepts one namespace, either provided as a CLI
argument or as an environment variable. Providing a list of namespaces
is not supported.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -160,6 +160,14 @@ pub(crate) fn build_cli() -> Command<'static> {
                 .required(false)
                 .help("Enable Sigstore verification [env: KUBEWARDEN_ENABLE_VERIFICATION=]"),
         )
+        .arg(
+            Arg::new("always-accept-admission-reviews-on-namespace")
+                .long("always-accept-admission-reviews-on-namespace")
+                .takes_value(true)
+                .env("ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE")
+                .required(false)
+                .help("Always accept AdmissionReviews that target the given namespace [env: ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE]"),
+        )
         .long_version(VERSION_AND_BUILTINS.as_str())
 }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -164,9 +164,9 @@ pub(crate) fn build_cli() -> Command<'static> {
             Arg::new("always-accept-admission-reviews-on-namespace")
                 .long("always-accept-admission-reviews-on-namespace")
                 .takes_value(true)
-                .env("ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE")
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -112,12 +112,17 @@ fn main() -> Result<()> {
     let (worker_pool_bootstrap_req_tx, worker_pool_bootstrap_req_rx) =
         oneshot::channel::<WorkerPoolBootRequest>();
 
+    let always_accept_admission_reviews_on_namespace = matches
```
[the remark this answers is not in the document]
viccuad: Normally I agree with declaring vars closer to where they are used. Was just a nit for consistency.

jvanz (approved):

src/cli.rs
```diff
@@ -160,6 +160,14 @@ pub(crate) fn build_cli() -> Command<'static> {
                 .required(false)
                 .help("Enable Sigstore verification [env: KUBEWARDEN_ENABLE_VERIFICATION=]"),
         )
+        .arg(
+            Arg::new("always-accept-admission-reviews-on-namespace")
+                .long("always-accept-admission-reviews-on-namespace")
+                .takes_value(true)
+                .env("ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE")
```
flavio, line 167: All our environment variables have the `KUBEWARDEN_` prefix:

```suggestion
                .env("KUBEWARDEN_ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE")
```

src/cli.rs
```diff
@@ -160,6 +160,14 @@ pub(crate) fn build_cli() -> Command<'static> {
...
+                .required(false)
+                .help("Always accept AdmissionReviews that target the given namespace [env: ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE]"),
```
flavio, line 169:
```suggestion
                .help("Always accept AdmissionReviews that target the given namespace [env: KUBEWARDEN_ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE]"),
```

flavio (approved):
LGTM, I just think we should name the environment variable in a different way, just to stay consistent with the naming convention we have already in place.

**IMPORTANT:** If we change the name of the env variable like I propose, we will have to update https://github.com/kubewarden/kubewarden-controller/pull/212

raulcabello (approved):
LGTM

PR #256, opened 2022-05-19
Sign policy-server image

### LINKED ISSUE #255: Feature Request: Sign policy-server image with cosign

Artifacts are not signed, therefore we can't verify if they come from a trusted source

### Solution you'd like

Sign policy-server image using keyless in github actions

raulcabello:
Use cosign keyless to sign the policy-server image

Tested in my [fork](https://github.com/raulcabello/policy-server/runs/6503089885?check_suite_focus=true) 

```
COSIGN_EXPERIMENTAL=1 cosign verify ghcr.io/raulcabello/policy-server:latest | jq
```

output:

```
...
      "Issuer": "https://token.actions.githubusercontent.com",
      "Subject": "https://github.com/raulcabello/policy-server/.github/workflows/container-image.yml@refs/heads/main"
...
```

Fix #255

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -66,6 +66,9 @@ jobs:
   build-container-image:
     name: Build container image
     runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      id-token: write
     needs:
      - build-x86_64
      - build-aarch64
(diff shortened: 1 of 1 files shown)


viccuad (approved):
LGTM! Should we add a `cosign verify` for each of the signs? It of course would not check for the signer, but at least we know the signature was done corectly and sent to the transparency log.

flavio (commented):
Signing against a `tag` is not considered a good practice. Can we get the digest of the image we just pushed, so that we can sign THAT digest?

raulcabello:
@flavio you are right. Done in https://github.com/kubewarden/policy-server/pull/256/commits/87fcf71eb100115e15ecce8dfb68af39dfa04958 , it can be checked [here](https://github.com/raulcabello/policy-server/runs/6523786681?check_suite_focus=true)

flavio (approved):
Awesome!

Issue #260, opened 2022-05-25
Non-mutating policies in monitor mode will attempt to mutate


jordojordo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a policy is in monitor mode and `mutating` is `false` - the policy will attempt to mutate a request, which in effect will reject the request.

### Expected Behavior

Policies that are in monitor mode should not attempt to mutate any requests.

### Steps To Reproduce

1. Install [kubewarden-defaults](https://github.com/kubewarden/helm-charts/tree/main/charts/kubewarden-defaults) with `mode: monitor` and `mutating: false`
2. Attempt to create a non-privileged pod
3. The creation will be rejected

### Environment

```markdown
- OS: Linux
- Architecture: amd64
- Kubernetes version: v1.22.9+rke2r2
```

### Anything else?

The pod that was rejected:

![monitor-rejected-privileged](https://user-images.githubusercontent.com/40806497/170285454-d3c71572-52ea-42df-b861-606176704eb1.png)

The policy yaml:

![monitor-policy-yaml](https://user-images.githubusercontent.com/40806497/170285554-9af99a4d-f85f-40c4-ba11-197ab9c7d8fc.png)

floriankoch:
@jvanz this is not fixed , i have a mutating policy in monitor mode, I got This error
`"The Namespace "xxx" is invalid: metadata.deletionTimestamp: Invalid value: 0001-01-01 00:00:00 +0000 UTC: field is immutable"`

I run v1.0.0, and i know that the metadata.deletionTimestamp field can not be changed, this is another bug @flavio  is looking into

flavio:
I think there are two issues there:

  * The Go SDK used by @floriankoch to produce the mutation request is affected by a bug - I'm looking into that
  * Policy Server should just ignore the patched object returned by a mutating policy when the policy is deployed in monitor mode

I'll reopen this issue, to have someone look at the 2nd point

jvanz:
> I think there are two issues there:
> 
>     * The Go SDK used by @floriankoch to produce the mutation request is affected by a bug - I'm looking into that
> 
>     * Policy Server should just ignore the patched object returned by a mutating policy when the policy is deployed in monitor mode
> 
> 
> I'll reopen this issue, to have someone look at the 2nd point

I'll take a look in the 2nd point.

jvanz:
[...]

PR #262, opened 2022-05-26
fix: policies in monitor mode still rejecting requests.

jvanz:
Policies running in monitor mode were still rejecting request if the policies mutate the request. This commit changes the policy-server code to always accept request in monitor mode. Regardless of the request of the evaluation for the income request.

Fixes https://github.com/kubewarden/policy-server/issues/260

## Test

1. Install [kubewarden-defaults](https://github.com/kubewarden/helm-charts/tree/main/charts/kubewarden-defaults) with mode: monitor and mutating: false
2. Attempt to create a non-privileged pod
3. The creation will be rejected

### DIFF
--- a/src/worker.rs
+++ b/src/worker.rs
@@ -173,25 +173,24 @@ impl Worker {
         allowed_to_mutate: bool,
         validation_response: ValidationResponse,
     ) -> ValidationResponse {
-        if validation_response.patch.is_some() && !allowed_to_mutate {
-            // Return early -- a policy not allowed to mutate tried to
-            // mutate --, we don't care about the policy mode at this
-            // point, this will be a problem in Protect mode, so
-            // return the error too if we are in Monitor mode.
-            return ValidationResponse {
-                allowed: false,
-                status: Some(ValidationResponseStatus {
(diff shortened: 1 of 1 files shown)


src/worker.rs
```diff
@@ -173,25 +173,28 @@ impl Worker {
...
-            };
-        };
         match policy_mode {
-            PolicyMode::Protect => validation_response,
+            PolicyMode::Protect => {
+                if validation_response.patch.is_some() && !allowed_to_mutate {
+                    // Return early -- a policy not allowed to mutate tried to
+                    // mutate --, we don't care about the policy mode at this
+                    // point, this will be a problem in Protect mode, so
+                    // return the error too if we are in Monitor mode.
```
flavio, line 182: nit update these lines because they mention the protect/monitor mode. Now this block of code is always in "protect" mode. I think you can even drop these lines to be honest...

flavio (approved):
Good catch, I left a minor comment

src/worker.rs
jvanz: Done

flavio (approved):
thanks for the fix done to the comment

Issue #239, opened 2022-04-28
Policy Server Readiness Probe Fails after some time


floriankoch:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

After some time (~140m) the Policy Server Readiness Probe fails 
Because there is no liveness probe, i have no idea if the application works

We have run the policy server with tracing enabled - no error in the log

 Warning  Unhealthy  52s (x92 over 157m)  kubelet  Readiness probe failed: Get "https://10.244.14.179:8443/readiness": dial tcp 10.244.14.179:8443: connect: connection refused

Latest Log Messages
2022-04-28T07: 42: 57.434789Z  INFO validation{host="policy-server-default-7dbfb95cb8-nwh29" policy_id="clusterwide-allow-pod-privileged-psp-policy" kind="Pod" kind_group="" kind_version="v1" name="redacted-84f9d67d94-ljv5k" namespace="redacted" operation="CREATE" request_uid="7ac29b7c-b2ee-4bb3-84d1-462de6bf612a" resource="pods" resource_group="" resource_version="v1" subresource=""
}:policy_eval: policy_server: :worker: policy evaluation (monitor mode) policy_id="clusterwide-allow-pod-privileged-psp-policy" allowed_to_mutate=false response="ValidationResponse { uid: \"7ac29b7c-b2ee-4bb3-84d1-462de6bf612a\", allowed: true, patch_type: None, patch: None, status: Some(ValidationResponseStatus { message: Some(\"\"), code: None }) }"
2022-04-28T07: 42: 57.434921Z DEBUG validation{host="policy-server-default-7dbfb95cb8-nwh29" policy_id="clusterwide-allow-pod-privileged-psp-policy" kind="Pod" kind_group="" kind_version="v1" name="redacted-84f9d67d94-ljv5k" namespace="redacted" operation="CREATE" request_uid="7ac29b7c-b2ee-4bb3-84d1-462de6bf612a" resource="pods" resource_group="" resource_version="v1" subresource="" allowed=true mutated=false
}: policy_server: :api: policy evaluated response="{\"apiVersion\":\"admission.k8s.io/v1\",\"kind\":\"AdmissionReview\",\"response\":{\"uid\":\"7ac29b7c-b2ee-4bb3-84d1-462de6bf612a\",\"allowed\":true}}"

### Expected Behavior

Policy Server Readiness does not fail without a reason

### Steps To Reproduce

Policy Server runs in a Azure AKS Cluster
[...]

PR #276, opened 2022-06-20
cleanup http server

flavio:
Prior to this commit, we used the low level hyper to create our HTTP server. The code being used was pretty complex due to the low level nature of this library.
Moreover, the code wasn't robus enough. In certain cases the HTTP server could fail and cause the whole policy-server to drop incoming requests.

This kind of failures was hard to reproduce, but some users have run into that.

I was able to reproduce that too with minikube. Malformed client TLS requests could cause the server to reject them and then enter an error state.

Instead of implementing all the possible workarounds for these kind of situations, this code now implements the HTTP and HTTPS server using the warp crate.

Warp is built on top of hyper, but provides a ready to be consumed high level API.
Our core business is not implementing HTTP(s) servers, hence by using this library we make our code more robust and improve its overall quality.

As a matter of fact, thanks to this change, a lot of obscure code and repetitive one has been dropped.
Also, a lot of top level dependencies have been removed, because they are now pulled via warp.

FIXES https://github.com/kubewarden/policy-server/issues/239

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.57"
+version = "1.0.58"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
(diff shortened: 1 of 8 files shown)


jvanz (approved):
LGTM. But I'll give a second round of review tomorrow. I'll try to run the code locally.

raulcabello (approved):
LGMT, Thanks! the code with warp is easier to understand! 
I run policy-server locally, did a few requests with curl and it is working as expected.

PR #282, opened 2022-07-08
update to latest policy evaluator

flavio:
- chore(deps): lock file maintenance (from https://github.com/kubewarden/policy-server/pull/281)
- Update to latest version of policy-evaluator to fix the compilation errors


flavio:
Once merged, I think we can tag a new release of policy-server.

viccuad (approved):

PR #285, opened 2022-07-11
[feat] allow text logs to not have color

flavio:
Allow user to have the default `text` log produce entries that are not colored.

This can be done in two ways:
  * Via the `--log-no-color` flag
  * By setting the following environment variable: `KUBEWARDEN_LOG_NO_COLOR=1`

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -57,6 +57,14 @@ pub(crate) fn build_cli() -> Command<'static> {
                 .help("Log output format"),
         )
         .arg(
+            Arg::new("log-no-color")
+                .long("log-no-color")
+                .takes_value(false)
+                .env("KUBEWARDEN_LOG_NO_COLOR")
+                .required(false)
+                .help("Disable colored output for logs [env: KUBEWARDEN_LOG_NO_COLOR=]"),
+        )
+        .arg(
             Arg::new("address")
                 .long("addr")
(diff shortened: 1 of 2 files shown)


flavio:
This is something requested by @floriankoch

raulcabello (approved):

viccuad (approved):
LGTM! just a nit, could we support [`NO_COLOR`](https://no-color.org/) instead?

flavio:
@viccuad: done

src/cli.rs
```diff
@@ -56,6 +56,14 @@ pub(crate) fn build_cli() -> Command<'static> {
                 .possible_values(&["text", "json", "otlp"])
                 .help("Log output format"),
         )
+        .arg(
+            Arg::new("log-no-color")
+                .long("log-no-color")
+                .takes_value(false)
+                .env("NO_COLOR")
```
jvanz, line 63: Should not be `KUBEWARDEN_LOG_NO_COLOR`?

flavio: I changed it to be `NO_COLOR` as @viccuad suggested. There's seems to be a consensus about different cli tools to have this specific flag. See here: https://no-color.org/

jvanz: Right. Sorry. I missed the @viccuad comment.

PR #287, opened 2022-07-13
Prometheus metrics fixes 

### LINKED ISSUE #263: Add the policy evaluation results in the tracing data of the policies running in monitor mode

### Current Behavior

While I was working in the [issue 260](https://github.com/kubewarden/policy-server/issues/260) I wondered if we should not change where we send the traces to Jaeger. Or at least, add other fields in the traces. [The traces sending to Jaeger happen after the ](https://github.com/kubewarden/policy-server/blob/main/src/worker.rs#L245-L290)`validation_response_with_constraints`[ call](https://github.com/kubewarden/policy-server/blob/main/src/worker.rs#L245-L290). This means that we lost the information about the actual policy evaluation.  In the end, we will always get accepted requests traces in the Jaeger. But we do not know if the request would be reject in the protected mode and the reason of that.

### Expected Behavior

The user should be able to see the actual policy evaluation results in the traces of policies running in monitor mode

### Steps To Reproduce

_No response_

### Environment

```markdown
Any
```

### Anything else?

_No response_

jvanz:
Updates the worker code used to run the policy evaluation to use the vanilla admission response from the policy in the metrics. Thus, the users will be able to actually see what policies are doing before the values updates in monitor and protect modes.

Remove the `resource_name` from the prometheus metrics to avoid too much time series, reducing the data stored and simplifying the query used to extract information from the metrics. This is a best practice:  https://prometheus.io/docs/practices/naming/

Fixes #263

### DIFF
--- a/src/metrics/mod.rs
+++ b/src/metrics/mod.rs
@@ -25,7 +25,6 @@ pub(crate) fn init_meter() -> metrics::Result<PushController> {
 pub struct PolicyEvaluation {
     pub(crate) policy_name: String,
     pub(crate) policy_mode: String,
-    pub(crate) resource_name: String,
     pub(crate) resource_kind: String,
     pub(crate) resource_namespace: Option<String>,
     pub(crate) resource_request_operation: String,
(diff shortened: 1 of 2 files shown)


flavio (approved):
LGTM, thanks for the fix

raulcabello (approved):
LGTM

viccuad (approved):

PR #303, opened 2022-08-10
 feat: use docker_credential crate instead of DockerConfig

### LINKED ISSUE #302: Remove DockerConfig and use docker_credential crate instead
Adapt the code and consume latest [changes](https://github.com/kubewarden/policy-fetcher/pull/111) from policy-fetcher where `docker_credential` crate is used.

raulcabello:
Fix #302


flavio (approved):
LGTM, have you tested that the whole `env` trick works fine?

raulcabello:
yes, I did some local testing and it's working fine

PR #308, opened 2022-08-15
Use builder container to build policy server.

### LINKED ISSUE #304: Release workflow  is flacky due connection issues with `musl`

### Current Behavior

From time to time, building the ARM release of Policy Server fails. The failure happens because there's an error downloading the musl release we use to perform the cross build.

This is really bad, because it doesn't allow us to release new features/fixes.

### Expected Behavior

First of all, the build should be stable. It should be green every time.

While addressing this issue, it would be nice (but not mandatory) to address also these current limitations:

  * We are downloading a musl release that is not versioned. This makes things not reproducible and less predictable.
  * When doing a SBOM, this would be problematic

Currently we are using a two stage build process. The 1st stage builds a statically linked binary for policy-server, which is then copied into a distroless container.

Maybe we can perform stage 1 inside some linux distribution container image which comes with the musl toolkit.

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- Architecture:
```

### Anything else?

_No response_

jvanz:
The people maintaining muscl.cc website have blocked IP ranges from some
cloud providers after a big spike of traffic from them. This affects
Kubewarden builds because it is not possible to download the toolchain
in the Github actions. To workaround that, a container image with the
Musl toolchains used to build the policy server has been created.

This builder container is used in the Github worflow to build the
Policy Server binaries as done before in VMs. The binaries than are
copied in the Policy Server container image as before.

Fix #304 

**NOTE**: While you review this PR, I'm evaluating if it's possible to use the musl available in the `ubuntu` and `debian` distros. Even if it is possible, afaics, this approach has a pro which is that the developers can build the binaries locally with no big configuration changes.


viccuad (commented):
The PR works as intended and is a valid approach. Still, looking forward to the other approach.
The other approach of cross-building with distro packages would have less maintenance, and simplify SBOM and SLSA too (I suppose we would need to track the builder image as we do with the rest).

jvanz:
> The PR works as intended and is a valid approach. Still, looking forward to the other approach. The other approach of cross-building with distro packages would have less maintenance, and simplify SBOM and SLSA too (I suppose we would need to track the builder image as we do with the rest).

I fear that will not be possible. The `musl` ubuntu packages are available in multiple architecture. But not  for cross compilation. Thus, we will need a aarch64 machine in that case. Which bring us back to the long build time that I would not like to see again. Thus, we have two options:

1. Create a package to be distributed via package manage like zypper, apt-get. We can use Open Build Service for that. 
2. Use container images. 

I prefer option 2. I believe that if we go with option 1 soon on later we will forget to update the package definition and will use some out of date version that would lead to other problems.

jvanz:
I've also tried to use the docker [container](https://hub.docker.com/layers/x86_64/muslcc/x86_64/aarch64-linux-musl/images/sha256-2d106ab72b2b5ea5ac7696a6b3d7c5f4f98d24f99230fb784d991c453034c017?context=explore) made available by the Musl team. But rustup does not work well there. Actually it cannot even install the toolchain due some issue with rustc. Afaics, conforming to the rust [platform support page](https://doc.rust-lang.org/nightly/rustc/platform-support.html), it's not guarantee that the tools will work in the devel set up that we are working on. So, we cannot use this option now.

viccuad:
I would prefer to not maintain an intermediate container image, that would need to be signed and SLSA checked eventually.

On Debian, it seems that [`musl-dev` package](https://packages.debian.org/bookworm/musl-dev) provides a gcc wrapper script for cross-compilation. E.g: `/usr/lib/aarch64-linux-musl/musl-gcc.specs` file in [the arm64 arch of the package](https://packages.debian.org/bookworm/arm64/musl-dev/filelist).
So maybe if installed with multiarch, and using the correct `gcc-*-cross` crosscompiler, it should work. 

This is in Debian, in Ubuntu there seems to be some differences, see [here](https://wiki.debian.org/ToolChain/Cross#Ubuntu_way) (also, that wiki page is quite extensive).

Another option could be https://github.com/GregorR/musl-cross or https://github.com/richfelker/musl-cross-make, to build the `musl` crosscompiler needed.

raulcabello:
I would prefer one of the options provided by @viccuad, so we wouldn't have to maintain an intermediate container image. 
However I would be happy with the approach of this PR if these options are not possible or take too much time to implement.

jvanz:
> I would prefer to not maintain an intermediate container image, that would need to be signed and SLSA checked eventually.

I believe that understand the concerns. But I do not think signing is an issue. Because we already do that. Furthermore,  as far as I can see,  using or not container will not make so much difference about SLSA. In the end, we will need to check SLSA no matter where the build happens. The current approach is using the standard Ubuntu binaries so, once we got the attestations to Ubuntu we should not be in big troubles. I have an uninformed feeling that the simple fact of using Musl have a potential to be a bigger headache comparing where we do the build. 

But this is me with zero experience with SLSA sharing some thoughts after a quick read in its documentations.

> 
> On Debian, it seems that [`musl-dev` package](https://packages.debian.org/bookworm/musl-dev) provides a gcc wrapper script for cross-compilation. E.g: `/usr/lib/aarch64-linux-musl/musl-gcc.specs` file in [the arm64 arch of the package](https://packages.debian.org/bookworm/arm64/musl-dev/filelist). So maybe if installed with multiarch, and using the correct `gcc-*-cross` crosscompiler, it should work.
> 
> This is in Debian, in Ubuntu there seems to be some differences, see [here](https://wiki.debian.org/ToolChain/Cross#Ubuntu_way) (also, that wiki page is quite extensive).
> 
> Another option could be https://github.com/GregorR/musl-cross or https://github.com/richfelker/musl-cross-make, to build the `musl` crosscompiler needed.

https://github.com/GregorR/musl-cross is for old GCC versions. So, I would not consider it. 
Let me me try again using these packages. 

> 
> A third option could be to cross-compile to arm in an Alpine image (which already uses musl).

This will add a new container image in the story, isn't it? Furthermore, aren't you describing  something that I've tried and described in a [previous comment](https://github.com/kubewarden/policy-server/pull/308#issuecomment-1217006351)? I've used the Musl container image with Musl already installed. But rustc does not work properly.

Dockerfile_builder
```diff
@@ -0,0 +1,22 @@
...
+RUN $HOME/.cargo/bin/rustup target add "${TARGETARCH}-unknown-linux-musl"
+
+COPY . /policy-server
+WORKDIR /policy-server
+
+RUN CC="/${TARGETARCH}-linux-musl-cross/bin/${TARGETARCH}-linux-musl-cc" \
+        LINKER="/${TARGETARCH}-linux-musl-cross/bin/${TARGETARCH}-linux-musl-ld" \
+	PATH="$PATH:/${TARGETARCH}-linux-musl-cross/bin" \
+	$HOME/.cargo/bin/cargo build --target="${TARGETARCH}-unknown-linux-musl" --release
+RUN mv "target/${TARGETARCH}-unknown-linux-musl/release/policy-server" policy-server
```
jvanz, line 22: This whole file can be merge with the old `Dockerfile` to remove one container image from the scenario. It will required some changes in the github workflow.

.github/workflows/container-image.yml
```diff
@@ -1,5 +1,7 @@
-name: build container image
+name: Build container image
 on:
+  # User can trigger this workflow in the Github UI
+  workflow_dispatch:
```
viccuad, line 4: I would drop this line, for other github workflows is ok, but for releases we shouldn't overwrite and re-release them.

.github/workflows/container-image.yml
```diff
@@ -9,69 +11,81 @@ jobs:
...
+          sudo sed -i "$ a deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports jammy-security main restricted" /etc/apt/sources.list
+          sudo sed -i "$ a deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports jammy-security universe" /etc/apt/sources.list
+          sudo sed -i "$ a deb [arch=arm64] http://ports.ubuntu.com/ubuntu-ports jammy-security multiverse" /etc/apt/sources.list
+
+          sudo apt update -y
+      - name: Install Musl and configure gcc spec
+        run: |
+          sudo apt install -y musl-dev:${{ matrix.arch }}
+          sudo sed -i 1d /usr/lib/${{ matrix.targetarch }}-linux-musl/musl-gcc.specs
+          sudo sed -i "s/*cpp_options/+cpp_options/g" /usr/lib/${{ matrix.targetarch }}-linux-musl/musl-gcc.specs
```
viccuad, line 53: I would like to add a comment before the 2 lines patching the spec. Something like:
`# patching the .spec file, as by default it has a bug where it tries to set old_cpp_options but it already exists. using *+cpp_options achieves the same desired functionality of appending preexisting options`

viccuad (commented):
Wooho! LGTM, just some comments

jvanz:
Running x86_64 binary:

```console
jvanz@earth:~/suse/kubewarden-end-to-end-tests> docker rm ps; docker create --name ps ghcr.io/jvanz/policy-server:latest
ps
6c3ed2aba11c44be4e9fabb773034e047c288934b3519ed3a6644a95ec986c26
jvanz@earth:~/suse/kubewarden-end-to-end-tests> docker cp ps:/policy-server ./policy-server
jvanz@earth:~/suse/kubewarden-end-to-end-tests> qemu-x86_64 ./policy-server -h
policy-server 1.1.1
Flavio Castelli <fcastelli@suse.com>:Rafael Fernández López <rfernandezlopez@suse.com>:Víctor
Cuadrado Juan <vcuadradojuan@suse.de>


USAGE:
    policy-server [OPTIONS]

OPTIONS:
        --addr <address>                                                                                 Bind against ADDRESS [env: KUBEWARDEN_BIND_ADDRESS=] [default: 0.0.0.0]
        --always-accept-admission-reviews-on-namespace <always-accept-admission-reviews-on-namespace>    Always accept AdmissionReviews that target the given namespace [env: KUBEWARDEN_ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE=]
        --cert-file <cert-file>                                                                          Path to an X.509 certificate file for HTTPS [env: KUBEWARDEN_CERT_FILE=] [default: ]
        --docker-config-json-path <docker-config-json-path>                                              Path to a Docker config.json-like path. Can be used to indicate registry authentication details [env: KUBEWARDEN_DOCKER_CONFIG_JSON_PATH=]
        --enable-metrics                                                                                 Enable metrics [env: KUBEWARDEN_ENABLE_METRICS=]
        --enable-verification                                                                            Enable Sigstore verification [env: KUBEWARDEN_ENABLE_VERIFICATION=]
    -h, --help                                                                                           Print help information
        --key-file <key-file>                                                                            Path to an X.509 private key file for HTTPS [env: KUBEWARDEN_KEY_FILE=] [default: ]
        --log-fmt <log-fmt>                                                                              Log output format [env: KUBEWARDEN_LOG_FMT=] [default: text] [possible values: text, json, otlp]
        --log-level <log-level>                                                                          Log level [env: KUBEWARDEN_LOG_LEVEL=] [default: info] [possible values: trace, debug, info, warn, error]
        --log-no-color                                                                                   Disable colored output for logs [env: NO_COLOR=]
        --policies <policies>                                                                            YAML file holding the policies to be loaded and their settings [env: KUBEWARDEN_POLICIES=] [default: policies.yml]
        --policies-download-dir <policies-download-dir>                                                  Download path for the policies [env: KUBEWARDEN_POLICIES_DOWNLOAD_DIR=] [default: .]
        --port <port>                                                                                    Listen on PORT [env: KUBEWARDEN_PORT=] [default: 3000]
        --sigstore-cache-dir <sigstore-cache-dir>                                                        Directory used to cache sigstore data [env: KUBEWARDEN_SIGSTORE_CACHE_DIR=] [default: sigstore-data]
        --sources-path <sources-path>                                                                    YAML file holding source information (https, registry insecure hosts, custom CA's...) [env: KUBEWARDEN_SOURCES_PATH=]
    -V, --version                                                                                        Print version information
        --verification-path <verification-path>                                                          YAML file holding verification information (URIs, keys, annotations...) [env: KUBEWARDEN_VERIFICATION_CONFIG_PATH=]
        --workers <workers>                                                                              Number of workers thread to create [env: KUBEWARDEN_WORKERS=]

```
Running aarch64 binary:
```console
jvanz@earth:~/suse/kubewarden-end-to-end-tests> docker rm ps; docker create --name ps  ghcr.io/jvanz/policy-server:latest@sha256:54aac49df6b9ad755496b396fa674c00eea4e9182374ca36a8c83fe2dbfc8b69
ps
Unable to find image 'ghcr.io/jvanz/policy-server:latest@sha256:54aac49df6b9ad755496b396fa674c00eea4e9182374ca36a8c83fe2dbfc8b69' locally
ghcr.io/jvanz/policy-server@sha256:54aac49df6b9ad755496b396fa674c00eea4e9182374ca36a8c83fe2dbfc8b69: Pulling from jvanz/policy-server
fe662701b57b: Already exists 
a6727b88e570: Already exists 
9f9c109645f1: Pull complete 
Digest: sha256:54aac49df6b9ad755496b396fa674c00eea4e9182374ca36a8c83fe2dbfc8b69
Status: Downloaded newer image for ghcr.io/jvanz/policy-server@sha256:54aac49df6b9ad755496b396fa674c00eea4e9182374ca36a8c83fe2dbfc8b69
WARNING: The requested image's platform (linux/arm64) does not match the detected host platform (linux/amd64) and no specific platform was requested
7a4d9da5c6277a8e5879cd5741ecf4712108c6e3eaac7cab809c8f8e035496ed
jvanz@earth:~/suse/kubewarden-end-to-end-tests> docker cp ps:/policy-server ./policy-server
jvanz@earth:~/suse/kubewarden-end-to-end-tests> qemu-aarch64 ./policy-server -h
policy-server 1.1.1
Flavio Castelli <fcastelli@suse.com>:Rafael Fernández López <rfernandezlopez@suse.com>:Víctor
Cuadrado Juan <vcuadradojuan@suse.de>


USAGE:
    policy-server [OPTIONS]

OPTIONS:
        --addr <address>                                                                                 Bind against ADDRESS [env: KUBEWARDEN_BIND_ADDRESS=] [default: 0.0.0.0]
        --always-accept-admission-reviews-on-namespace <always-accept-admission-reviews-on-namespace>    Always accept AdmissionReviews that target the given namespace [env: KUBEWARDEN_ALWAYS_ACCEPT_ADMISSION_REVIEWS_ON_NAMESPACE=]
        --cert-file <cert-file>                                                                          Path to an X.509 certificate file for HTTPS [env: KUBEWARDEN_CERT_FILE=] [default: ]
        --docker-config-json-path <docker-config-json-path>                                              Path to a Docker config.json-like path. Can be used to indicate registry authentication details [env: KUBEWARDEN_DOCKER_CONFIG_JSON_PATH=]
        --enable-metrics                                                                                 Enable metrics [env: KUBEWARDEN_ENABLE_METRICS=]
        --enable-verification                                                                            Enable Sigstore verification [env: KUBEWARDEN_ENABLE_VERIFICATION=]
    -h, --help                                                                                           Print help information
        --key-file <key-file>                                                                            Path to an X.509 private key file for HTTPS [env: KUBEWARDEN_KEY_FILE=] [default: ]
        --log-fmt <log-fmt>                                                                              Log output format [env: KUBEWARDEN_LOG_FMT=] [default: text] [possible values: text, json, otlp]
        --log-level <log-level>                                                                          Log level [env: KUBEWARDEN_LOG_LEVEL=] [default: info] [possible values: trace, debug, info, warn, error]
        --log-no-color                                                                                   Disable colored output for logs [env: NO_COLOR=]
        --policies <policies>                                                                            YAML file holding the policies to be loaded and their settings [env: KUBEWARDEN_POLICIES=] [default: policies.yml]
        --policies-download-dir <policies-download-dir>                                                  Download path for the policies [env: KUBEWARDEN_POLICIES_DOWNLOAD_DIR=] [default: .]
        --port <port>                                                                                    Listen on PORT [env: KUBEWARDEN_PORT=] [default: 3000]
        --sigstore-cache-dir <sigstore-cache-dir>                                                        Directory used to cache sigstore data [env: KUBEWARDEN_SIGSTORE_CACHE_DIR=] [default: sigstore-data]
        --sources-path <sources-path>                                                                    YAML file holding source information (https, registry insecure hosts, custom CA's...) [env: KUBEWARDEN_SOURCES_PATH=]
    -V, --version                                                                                        Print version information
        --verification-path <verification-path>                                                          YAML file holding verification information (URIs, keys, annotations...) [env: KUBEWARDEN_VERIFICATION_CONFIG_PATH=]
        --workers <workers>                                                                              Number of workers thread to create [env: KUBEWARDEN_WORKERS=]

```

.github/workflows/container-image.yml
viccuad: fixed in https://github.com/kubewarden/policy-server/pull/308/commits/729c17266e0c7180f26724142359f5daed27b6ae

viccuad:
merging, to let it build the `:latest` tag! wooho!

PR #312, opened 2022-08-25
SBOM file generation.

### LINKED ISSUE #305: policy-server: create SBOM file
Each release should have a SBOM file that allows to track the origin of all the dependencies.

jvanz:
Adds new steps in the build and release workflow to generate, sign and publish a SBOM file for Policy Server.

Fix #305

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -1,4 +1,4 @@
-name: Build container image
+name: Build policy server
 on:
   push:
     branches:
@@ -28,6 +28,9 @@ jobs:
     needs:
       - ci
     steps:
+      - name: Install cosign
+        uses: sigstore/cosign-installer@main
+
       - name: Configure Ubuntu repositories
         run: |
           sudo dpkg --add-architecture arm64
@@ -82,11 +85,36 @@ jobs:
           cargo build --release --target ${{ matrix.targetarch }}-unknown-linux-musl
           mv target/${{ matrix.targetarch }}-unknown-linux-musl/release/policy-server policy-server-${{ matrix.targetarch }}
 
-      - name: Upload policy-server
+      - name: Generate SBOM
+        run: |
+          make download-spdx-sbom-generator sbom
+          # SBOM files should have "sbom" in the name due the CLO monitor
+          # https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom
+          mv bom-cargo.json policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+
+      - name: Sign BOM file
+        run: |
+          cosign sign-blob --output-certificate policy-server-${{ matrix.targetarch }}-sbom.spdx.cert \
+            --output-signature policy-server-${{ matrix.targetarch }}-sbom.spdx.sig \
+            policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+        env:
+          COSIGN_EXPERIMENTAL: 1
+
+      - name: Upload policy-server binary
         uses: actions/upload-artifact@v2
         with:
           name: policy-server-${{ matrix.targetarch }}
-          path: policy-server-${{ matrix.targetarch }}
+          path: |
+            policy-server-${{ matrix.targetarch }}
+
+      - name: Upload policy-server SBOM files
+        uses: actions/upload-artifact@v2
+        with:
+          name: policy-server-${{ matrix.targetarch }}-sbom
+          path: |
+            policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+            policy-server-${{ matrix.targetarch }}-sbom.spdx.cert
+            policy-server-${{ matrix.targetarch }}-sbom.spdx.sig
 
   build-container-image:
     name: Build policy server container image
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /target
+bom-cargo.json
+/bin
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,9 +2,11 @@
 name = "policy-server"
 version = "1.1.2"
 authors = [
+  "Kubewarden Developers <kubewarden@suse.de>",
   "Flavio Castelli <fcastelli@suse.com>",
   "Rafael Fernández López <rfernandezlopez@suse.com>",
-  "Víctor Cuadrado Juan <vcuadradojuan@suse.de>"
+  "Víctor Cuadrado Juan <vcuadradojuan@suse.de>",
+  "José Guilherme Vanz <jguilhermevanz@suse.com>"
 ]
 edition = "2018"
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
-HYPERFINE := $(shell command -v hyperfine 2> /dev/null)
 IMG ?= policy-server:latest
+BINDIR ?= bin
+SBOM_GENERATOR_TOOL_VERSION ?= v0.0.15
 
 .PHONY: build
 build:
@@ -31,3 +32,16 @@ tag:
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+
+bin:
+	mkdir $(BINDIR)
+
+.PHONY: download-spdx-sbom-generator
+download-spdx-sbom-generator: bin
+	curl -L -o $(BINDIR)/spdx-sbom-generator-$(SBOM_GENERATOR_TOOL_VERSION)-linux-amd64.tar.gz https://github.com/opensbom-generator/spdx-sbom-generator/releases/download/$(SBOM_GENERATOR_TOOL_VERSION)/spdx-sbom-generator-$(SBOM_GENERATOR_TOOL_VERSION)-linux-amd64.tar.gz
+	tar -xf ./$(BINDIR)/spdx-sbom-generator-$(SBOM_GENERATOR_TOOL_VERSION)-linux-amd64.tar.gz --directory $(BINDIR)
+
+
+.PHONY: sbom
+sbom:
+	./$(BINDIR)/spdx-sbom-generator -f json
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,23 +1,49 @@
-name: policy-server release
+name: Release policy server
 on:
-  push:
-    tags:
-    - 'v*'
+  workflow_run:
+    workflows: ["Build policy server"]
+    types:
+      - completed
+    branches:
+      - "v*"
 jobs:
-  ci:
-    uses: kubewarden/policy-server/.github/workflows/tests.yml@main
   release:
     name: Create release
     runs-on: ubuntu-latest
-    needs:
-      - ci
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     steps:
+      - name: Download build artifacts
+        uses: actions/github-script@v6
+        with:
+          script: |
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: context.payload.workflow_run.id,
+            });
+            const matchArtifact = allArtifacts.data.artifacts.filter((artifact) => {
+              return artifact.name.startsWith("policy-server")
+            });
+            for (const artifact of matchArtifact) {
+              let download = await github.rest.actions.downloadArtifact({
+                 owner: context.repo.owner,
+                 repo: context.repo.repo,
+                 artifact_id: artifact.id,
+                 archive_format: 'zip',
+              });
+              let fs = require('fs');
+              file_path = `${process.env.GITHUB_WORKSPACE}/${artifact.name}.zip`;
+              fs.writeFileSync(file_path, Buffer.from(download.data));
+            }
+
       - name: Create release
-        uses: actions/create-release@v1
+        uses: softprops/action-gh-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ github.ref }}
-          release_name: Release policy-server ${{ github.ref }}
+          tag_name: ${{ github.event.workflow_run.head_branch }}
+          name: Release policy-server ${{ github.event.workflow_run.head_branch }}
           draft: false
-          prerelease: ${{ contains(github.ref, '-alpha') || contains(github.ref, '-beta') || contains(github.ref, '-rc') }}
+          prerelease: ${{ contains(github.event.workflow_run.head_branch, '-alpha') || contains(github.event.workflow_run.head_branch, '-beta') || contains(github.event.workflow_run.head_branch, '-rc') }}
+          files: |
+            policy-server*
--- a/README.md
+++ b/README.md
@@ -119,3 +119,10 @@ Alternatively, the `policy-server` binary can be built in this way:
 ```shell
 $ make build
 ```
+
+# Software bill of materials
+
+Policy server has its software bill of materials (SBOM) published every release.
+It follows the [SPDX](https://spdx.dev/) version 2.2 format and it can be found
+together with the signature and certificate used to signed it in the
+[release assets](https://github.com/kubewarden/policy-server/releases)


jvanz:
@viccuad , in order to have a more complete SBOM file, the tool needs access to the files generated by the build. Thus, the release workflow needs to have access to the target directory. That's the reason I've merged the release worflows into a single file. I can split it again and call the same worflows to build the binaries in both workflows (`ci` and `release`). Do you prefer that? One cons of this approach is that the build will run twice when a tag is released. What do you think?

Makefile
```diff
@@ -31,3 +31,21 @@ tag:
...
+	./sbom-tool generate \
+		-D true \
+		-V Verbose \
+		-b ./target/release \
+		-bc . \
+		-m . \
+		-di ${IMG} \
+		-nsb https://kubewarden.io \
+		-nsu policy-server \
+		-pn policy-server \
```
viccuad, line 50: I think we should add `-ps` (package supplier), set to Kubewarden or `https://kubewarden.io.

Makefile
```diff
@@ -31,3 +31,21 @@ tag:
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+
+sbom-tool:
+	curl -Lo sbom-tool https://github.com/microsoft/sbom-tool/releases/latest/download/sbom-tool-linux-x64
```
viccuad, line 36: A pity that they don't sign linux binaries yet...

.github/workflows/container-image.yml
```diff
@@ -1,4 +1,4 @@
-name: Build container image
+name: Build and release policy server
```
viccuad, line 1: (opening thread here randomly)

> @viccuad , in order to have a more complete SBOM file, the tool needs access to the files generated by the build. Thus, the release workflow needs to have access to the target directory. That's the reason I've merged the release worflows into a single file. I can split it again and call the same worflows to build the binaries in both workflows (ci and release). Do you prefer that? One cons of this approach is that the build will run twice when a tag is released. What do you think?

But, with that, it means that the release job is run twice, possibly making 2 releases, or making 1 and then overwriting. I think we shouldn't do this, releases should be built reproducibly and it needs to be possible to automatically subscribe and consume then nicely (instead of getting 2 notifications for them per 1 real release).

As I understand, the SBOM creation needs to happen in the build job. I would like to have 1 workflow that:
1. builds the binary, generates the SBOM, signs both.
2. builds the container image, generates SBOM of container image (maybe with [`docker sbom`](https://docs.docker.com/engine/sbom/)), signs both.

And then the release workflow that we have, where we just download all the previous artifacts, and create a new GH release if there's a tag.

Makefile
```diff
@@ -31,3 +31,21 @@ tag:
...
+	chmod +x sbom-tool
+
+.PHONY: sbom
+sbom: sbom-tool
+	./sbom-tool generate \
```
viccuad, line 41: I wonder if we could get the output files to be neatly put in `./target/release/` or similar.. but doesn't really matter. Also, `sbom-tool` help falls a bit short...

Makefile
jvanz: Yes, but I think it's on the way: https://github.com/microsoft/sbom-tool/issues/44

Makefile
jvanz: This is an option not released it: https://github.com/microsoft/sbom-tool/commit/66fe849ce701998264da7f9cd731f937500589eb

viccuad: ooh, then let's skip it.

.github/workflows/container-image.yml
jvanz: We will not get additional notification. We will continue to have two as the current version. One for the job witch creates the release and another for the job witch build the container image. 

What will run twice is the container build process. Once for the `latest` tag and another for the `v*` tag. Each one in a different job. Witch will consume more resources and time and I would like to avoid. But at the time of the written I could not find a easy way to share artifacts between two different jobs as our current version is. I do not want to add to much complicated steps to make the communication works, keeping it simple and generating always the same results. Even if the push multiple commits in to main while releasing a tag. 

Let me see if I can find something else... 

PS: The current version of this PR already have some changes testing some changes since the first comment that originate this discussion.

flavio (commented):
I don't understand if the final SPDX file will contain a list of all the dependencies of the `policy-server` binary (basically the contents of `Cargo.lock`).

I tried running the code locally and it failed:

```console
[INFO] Detection time: 1.7895835 seconds. 
[INFO] Scan Manifest file: /tmp/ScanManifest_20220830184522.json 
##[debug]Using the DirectoryTraversingFileToJsonProvider provider for the files workflow.
##[debug]Enumerating files under the root path ./target/release.
##[debug]Splitting the workflow into 8 threads.
##[debug]Running the generation workflow ...
##[debug]Using the CGScannedPackagesProvider provider for the packages workflow.
##[debug]Scanning for packages under the root path ..
##[debug]Splitting the workflow into 8 threads.
##[debug]Running the generation workflow ...
##[debug]Using cached CD scan result for the call with the same arguments
##[debug]Wrote 11 package elements in the SBOM.
##[debug]Found value for header PackageName in internal metadata.
##[error]Encountered an error while generating the manifest.
##[error]Error details: Unable to generate a package version based on provided parameters. Please provide the package version in the 'PackageVersion' parameter.
##[information]Finished execution of the Generate workflow SBOMTelemetry {Result=Failure, Errors=ErrorContainer`1 {Count=0, Errors=[]}, Parameters=Configuration {BuildDropPath=ConfigurationSetting`1 {Value="./target/release", Source=CommandLine, IsDefaultSource=False}, BuildComponentPath=ConfigurationSetting`1 {Value=".", Source=CommandLine, IsDefaultSource=False}, BuildListFile=null, ManifestPath=null, ManifestDirPath=ConfigurationSetting`1 {Value="./_manifest", Source=CommandLine, IsDefaultSource=False}, OutputPath=null, Parallelism=ConfigurationSetting`1 {Value=8, Source=Default, IsDefaultSource=True}, Verbosity=ConfigurationSetting`1 {Value=Verbose, Source=CommandLine, IsDefaultSource=False}, ConfigFilePath=null, ManifestInfo=ConfigurationSetting`1 {Value=null, Source=Default, IsDefaultSource=True}, HashAlgorithm=ConfigurationSetting`1 {Value=AlgorithmName {Name="SHA256"}, Source=Default, IsDefaultSource=True}, RootPathFilter=null, CatalogFilePath=null, ValidateSignature=ConfigurationSetting`1 {Value=False, Source=Default, IsDefaultSource=True}, IgnoreMissing=ConfigurationSetting`1 {Value=False, Source=Default, IsDefaultSource=True}, ManifestToolAction=Generate, PackageName=ConfigurationSetting`1 {Value="policy-server", Source=CommandLine, IsDefaultSource=False}, PackageVersion=null, FilesList=null, PackagesList=null, TelemetryFilePath=null, DockerImagesToScan=ConfigurationSetting`1 {Value="policy-server:latest", Source=CommandLine, IsDefaultSource=False}, ExternalDocumentReferenceListFile=null, AdditionalComponentDetectorArgs=null, NamespaceUriUniquePart=ConfigurationSetting`1 {Value="policy-server", Source=CommandLine, IsDefaultSource=False}, NamespaceUriBase=ConfigurationSetting`1 {Value="https://kubewarden.io", Source=CommandLine, IsDefaultSource=False}, GenerationTimestamp=null, FollowSymlinks=ConfigurationSetting`1 {Value=True, Source=Default, IsDefaultSource=True}, DeleteManifestDirIfPresent=ConfigurationSetting`1 {Value=True, Source=CommandLine, IsDefaultSource=False}}, SBOMFormatsUsed=[SBOMFile {SbomFormatName=ManifestInfo {Name="SPDX", Version="2.2"}, SbomFilePath="./_manifest/spdx_2.2/manifest.spdx.json", FileSizeInBytes=4189034}], Timings=[Timing {EventName="Packages generation time", TimeSpan="00:00:00.0348830"}, Timing {EventName="Files generation time", TimeSpan="00:00:06.4819372"}, Timing {EventName="Total generation time", TimeSpan="00:00:06.5880395"}], Switches={}, Exceptions={["System.ArgumentException"]="Unable to generate a package version based on provided parameters. Please provide the package version in the 'PackageVersion' parameter."}}
make: *** [Makefile:41: sbom] Error 1
```

jvanz:
> I don't understand if the final SPDX file will contain a list of all the dependencies of the `policy-server` binary (basically the contents of `Cargo.lock`).
> 
> I tried running the code locally and it failed:
> 
> ```
> [INFO] Detection time: 1.7895835 seconds. 
> [INFO] Scan Manifest file: /tmp/ScanManifest_20220830184522.json 
> ##[debug]Using the DirectoryTraversingFileToJsonProvider provider for the files workflow.
> ##[debug]Enumerating files under the root path ./target/release.
> ##[debug]Splitting the workflow into 8 threads.
> ##[debug]Running the generation workflow ...
> ##[debug]Using the CGScannedPackagesProvider provider for the packages workflow.
> ##[debug]Scanning for packages under the root path ..
> ##[debug]Splitting the workflow into 8 threads.
> ##[debug]Running the generation workflow ...
> ##[debug]Using cached CD scan result for the call with the same arguments
> ##[debug]Wrote 11 package elements in the SBOM.
> ##[debug]Found value for header PackageName in internal metadata.
> ##[error]Encountered an error while generating the manifest.
> ##[error]Error details: Unable to generate a package version based on provided parameters. Please provide the package version in the 'PackageVersion' parameter.
> ##[information]Finished execution of the Generate workflow SBOMTelemetry {Result=Failure, Errors=ErrorContainer`1 {Count=0, Errors=[]}, Parameters=Configuration {BuildDropPath=ConfigurationSetting`1 {Value="./target/release", Source=CommandLine, IsDefaultSource=False}, BuildComponentPath=ConfigurationSetting`1 {Value=".", Source=CommandLine, IsDefaultSource=False}, BuildListFile=null, ManifestPath=null, ManifestDirPath=ConfigurationSetting`1 {Value="./_manifest", Source=CommandLine, IsDefaultSource=False}, OutputPath=null, Parallelism=ConfigurationSetting`1 {Value=8, Source=Default, IsDefaultSource=True}, Verbosity=ConfigurationSetting`1 {Value=Verbose, Source=CommandLine, IsDefaultSource=False}, ConfigFilePath=null, ManifestInfo=ConfigurationSetting`1 {Value=null, Source=Default, IsDefaultSource=True}, HashAlgorithm=ConfigurationSetting`1 {Value=AlgorithmName {Name="SHA256"}, Source=Default, IsDefaultSource=True}, RootPathFilter=null, CatalogFilePath=null, ValidateSignature=ConfigurationSetting`1 {Value=False, Source=Default, IsDefaultSource=True}, IgnoreMissing=ConfigurationSetting`1 {Value=False, Source=Default, IsDefaultSource=True}, ManifestToolAction=Generate, PackageName=ConfigurationSetting`1 {Value="policy-server", Source=CommandLine, IsDefaultSource=False}, PackageVersion=null, FilesList=null, PackagesList=null, TelemetryFilePath=null, DockerImagesToScan=ConfigurationSetting`1 {Value="policy-server:latest", Source=CommandLine, IsDefaultSource=False}, ExternalDocumentReferenceListFile=null, AdditionalComponentDetectorArgs=null, NamespaceUriUniquePart=ConfigurationSetting`1 {Value="policy-server", Source=CommandLine, IsDefaultSource=False}, NamespaceUriBase=ConfigurationSetting`1 {Value="https://kubewarden.io", Source=CommandLine, IsDefaultSource=False}, GenerationTimestamp=null, FollowSymlinks=ConfigurationSetting`1 {Value=True, Source=Default, IsDefaultSource=True}, DeleteManifestDirIfPresent=ConfigurationSetting`1 {Value=True, Source=CommandLine, IsDefaultSource=False}}, SBOMFormatsUsed=[SBOMFile {SbomFormatName=ManifestInfo {Name="SPDX", Version="2.2"}, SbomFilePath="./_manifest/spdx_2.2/manifest.spdx.json", FileSizeInBytes=4189034}], Timings=[Timing {EventName="Packages generation time", TimeSpan="00:00:00.0348830"}, Timing {EventName="Files generation time", TimeSpan="00:00:06.4819372"}, Timing {EventName="Total generation time", TimeSpan="00:00:06.5880395"}], Switches={}, Exceptions={["System.ArgumentException"]="Unable to generate a package version based on provided parameters. Please provide the package version in the 'PackageVersion' parameter."}}
> make: *** [Makefile:41: sbom] Error 1
> ```

You need to define the version used in the file. To run locally, try: `TAG=v9.9.9 make sbom`.

jvanz:
> I don't understand if the final SPDX file will contain a list of all the dependencies of the `policy-server` binary (basically the contents of `Cargo.lock`).

As far as I could understand reading the spdx specification the file contains anything used to build the binary. Therefore, you can see all the files include in a package, the relationship between packages, checksum of everything, license and etc.

.github/workflows/container-image.yml
jvanz: @viccuad I've update the workflows to keep the same approach that we have now. Please, review it again.

flavio:
Thanks @jvanz, I was able to run the command locally.

I've looked at the `_manifest/spdx_2.2/manifest.spdx.json` file generated and I wonder how useful it would be.

For example, take a look at this snippet about the `addr2line` dependency:

```json
    {
      "fileName": "./deps/addr2line-28aec851f2e54f6a.d",
      "SPDXID": "SPDXRef-File--deps-addr2line-28aec851f2e54f6a.d-19D14F2B41953D84DFBB935FED640F81D0809DFA",
      "checksums": [
        {
          "algorithm": "SHA256",
          "checksumValue": "4512d8e3dd2ea20f9b30e414254f9f04dcd8991d764b6f1a6d6b47abab74e6d2"
        },
        {
          "algorithm": "SHA1",
          "checksumValue": "19d14f2b41953d84dfbb935fed640f81d0809dfa"
        }
      ],
      "licenseConcluded": "NOASSERTION",
      "licenseInfoInFiles": [
        "NOASSERTION"
      ],
      "copyrightText": "NOASSERTION"
    },
```

This snippet doesn't tell me where the dependency is coming (crates.io release, a github repository,...) and it doesn't tell me the version being used. I can see what I assume is a git commit shasum inside of the folder  name `./deps/addr2line-28aec851f2e54f6a.d`.

This isn't what I expected to find inside of a SPDX report.

I did some quick search, and I ran [opensbom-generator/spdx-sbom-generator](https://github.com/opensbom-generator/spdx-sbom-generator) in this way:

```console
$ spdx-sbom-generator
```

This produced a `bom-cargo.spdx` which has contents that are more aligned with what I was expecting. For example, taking the same `addr2line` dependency:

```
##### Package representing the addr2line

PackageName: addr2line
SPDXID: SPDXRef-Package-addr2line-0.17.0
PackageVersion: 0.17.0
PackageSupplier: NOASSERTION
PackageDownloadLocation: https://github.com/gimli-rs/addr2line
FilesAnalyzed: false
PackageChecksum: SHA1: 59f51ecbe91443fc78708b82c2ad39645cba5d9d
PackageHomePage: https://github.com/rust-lang/crates.io-index
PackageLicenseConcluded: NOASSERTION
PackageLicenseDeclared: NOASSERTION
PackageCopyrightText: NOASSERTION
PackageLicenseComments: NOASSERTION
PackageComment: NOASSERTION

Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-rustc-demangle-0.1.21 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-smallvec-1.9.0 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-backtrace-0.3.66 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-clap-3.2.17 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-cpp-demangle-0.3.5 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-fallible-iterator-0.2.0 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-gimli-0.26.2 
Relationship: SPDXRef-Package-addr2line-0.17.0 DEPENDS_ON SPDXRef-Package-object-0.29.0 
```

This looks more "human-friendly".

Now, I'm not suggesting we have to move to this other tool. I just wonder if we are not invoking the Microsoft one with the right set of parameters.

Makefile
viccuad: It seems they have just released a new version, and it needs `-ps kubewarden` now, or:
`Encountered error while running ManifestTool generation workflow. Error: The value of PackageSupplier can't be null or empty.`

Makefile
viccuad, line 36: I think we should pin the version of the sbom-tool.

viccuad:
@flavio here, the `sbom-tool` outputs the expected contents for `addr2line` (in fact, the same you pasted). After applying https://github.com/kubewarden/policy-server/pull/312#discussion_r959387935.

.github/workflows/container-image.yml
viccuad: Thanks! that looks more streamlined for me, resolving!

Makefile
viccuad: also, we should download it to a `bin/` folder, and call it from there.

jvanz:
@viccuad @flavio , I've updated the PR to use the [spdx-sbom-generator](https://github.com/opensbom-generator/spdx-sbom-generator) tool and generate the right packages dependencies. 

Furthermore, I've also open an [issue](https://github.com/microsoft/sbom-tool/issues/126) in the `sbom-tool` to know why we cannot get the right packages dependencies relationships from cargo file.

.gitignore
```diff
@@ -1 +1,4 @@
-/target
\ No newline at end of file
+/target
+spdx-sbom-generator*
```
flavio, line 2: I think you can remove this line, because the binary is going to be downloaded under `/bin`, which is already ignored

.gitignore
```diff
@@ -1 +1,4 @@
...
+bom-cargo.json
+bin
```
flavio, line 4:
```suggestion
/bin
```

Makefile
```diff
@@ -31,3 +31,15 @@ tag:
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+
+bin:
+	mkdir bin
+
+spdx-sbom-generator: bin
```
flavio, line 38: This is always downloading the binary, even when it already exists. I wonder if we should simply assume the binary is already installed on the system and just download it via GitHub Actions...

.github/workflows/container-image.yml
```diff
@@ -82,11 +85,36 @@ jobs:
           cargo build --release --target ${{ matrix.targetarch }}-unknown-linux-musl
           mv target/${{ matrix.targetarch }}-unknown-linux-musl/release/policy-server policy-server-${{ matrix.targetarch }}
 
-      - name: Upload policy-server
+      - name: Generate SBOM
+        run: |
+          make sbom
+          # SBOM files should have "sbom" in the name due the CLO monitor
+          # https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom
+          cp bom-cargo.json policy-server-${{ matrix.targetarch }}-sbom.spdx.json
```
flavio, line 93: I personally prefer to move it:

```suggestion
          mv bom-cargo.json policy-server-${{ matrix.targetarch }}-sbom.spdx.json
```

Cargo.toml
```diff
@@ -2,9 +2,11 @@
 name = "policy-server"
 version = "1.1.2"
 authors = [
+  "Kubewarden",
```
viccuad, line 5:
```suggestion
  "Kubewarden Developers <kubewarden@suse.de>"
```

Would match what we have in `kwctl`.Maybe something like `Kubewarden Developers (github @kubewarden/kubewarden-developers)` would fit more, but it seems that[ spdx expects only emails there](https://spdx.github.io/spdx-spec/v2.3/package-information/#763-examples)

Makefile
jvanz: I've removed the dependencies between the targets and leave the target to download the tool to run by the users.

.github/workflows/container-image.yml
```diff
@@ -82,11 +85,36 @@ jobs:
...
+
+      - name: Sign BOM file
+        run: |
+          cosign sign-blob --output-certificate policy-server-${{ matrix.targetarch }}-sbom.spdx.cert \
+            --output-signature policy-server-${{ matrix.targetarch }}-sbom.spdx.sig \
+            policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+        env:
+          COSIGN_EXPERIMENTAL: 1
+
+      - name: Upload policy-server binary
```
viccuad, line 103: maybe we could sign the binary and upload the binary and its signature to the interim artifacts.
The signatures could be checked when building the container image. I suppose this may be needed for SLSA.
Not much improvement for our current security stance, as we depend on GHA and artifacts anyway. But in the future we could build in our own worker.

viccuad (approved):
Approving, I'm happy to merge as it is.

We could add SBOM for both build image and resulting container image in following PRs.

.github/workflows/container-image.yml
flavio: Yes, I think this would be a nice improvement. However I would not block this PR to get this implemented, I would rather file a new issue

Makefile
```diff
@@ -31,3 +32,16 @@ tag:
...
+bin:
+	mkdir $(BINDIR)
+
+.PHONY: download-spdx-sbom-generator
+download-spdx-sbom-generator: bin
+	curl -L -o $(BINDIR)/spdx-sbom-generator-v0.0.15-linux-amd64.tar.gz https://github.com/opensbom-generator/spdx-sbom-generator/releases/download/v0.0.15/spdx-sbom-generator-v0.0.15-linux-amd64.tar.gz
+	tar -xf ./$(BINDIR)/spdx-sbom-generator-v0.0.15-linux-amd64.tar.gz --directory $(BINDIR)
+
+
+.PHONY: sbom-generator
```
flavio, line 45: nitpicking, I think this line is not needed/wrong

Makefile
```diff
@@ -1,5 +1,6 @@
 HYPERFINE := $(shell command -v hyperfine 2> /dev/null)
```
flavio, line 1: not related with this PR, but this could be probably dropped

Makefile
```diff
@@ -31,3 +32,16 @@ tag:
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+
+bin:
+	mkdir $(BINDIR)
+
+.PHONY: download-spdx-sbom-generator
+download-spdx-sbom-generator: bin
+	curl -L -o $(BINDIR)/spdx-sbom-generator-v0.0.15-linux-amd64.tar.gz https://github.com/opensbom-generator/spdx-sbom-generator/releases/download/v0.0.15/spdx-sbom-generator-v0.0.15-linux-amd64.tar.gz
```
flavio, line 41: would it make sense to have the spdx-bom-generator version abstracted via a dedicated variable?

flavio (approved):
LGTM, I left some minor comments that are not mandatory to get this PR merged

Makefile
```diff
@@ -31,3 +32,16 @@ tag:
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+
+bin:
+	mkdir $(BINDIR)
+
+.PHONY: download-spdx-sbom-generator
+download-spdx-sbom-generator: bin
```
viccuad, line 40:
```suggestion
download-spdx-sbom-generator: $(BINDIR)
```

.github/workflows/container-image.yml
jvanz: That's sound a good idea indeed. I've just created an issue to keep track of this improvement.

Makefile
jvanz: Removed.

Makefile
jvanz: Fixed.

Makefile
jvanz: Done

PR #320, opened 2022-09-06
Container image SBOM

### LINKED ISSUE #314: SBOM files for container
Currently we are generating the SBOM files for the policy server binaries. To improve even further our security and SSC, It's necessary to generate SBOM file to the container images generated during the release process and attach it as a release asset.

jvanz:
## Description

Adds steps in the container image release process to generate the SBOM file for it and upload it to the build artifacts and as a release asset. 

Fix #314

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -90,13 +90,14 @@ jobs:
           make download-spdx-sbom-generator sbom
           # SBOM files should have "sbom" in the name due the CLO monitor
           # https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom
-          mv bom-cargo.json policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+          mkdir policy-server-${{ matrix.targetarch }}-sbom
+          mv bom-cargo.json policy-server-${{ matrix.targetarch }}-sbom/policy-server-${{ matrix.targetarch }}-sbom.spdx.json
 
       - name: Sign BOM file
         run: |
-          cosign sign-blob --output-certificate policy-server-${{ matrix.targetarch }}-sbom.spdx.cert \
-            --output-signature policy-server-${{ matrix.targetarch }}-sbom.spdx.sig \
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+          cosign sign-blob --output-certificate policy-server-${{ matrix.targetarch }}-sbom/policy-server-${{ matrix.targetarch }}-sbom.spdx.cert \
+            --output-signature policy-server-${{ matrix.targetarch }}-sbom/policy-server-${{ matrix.targetarch }}-sbom.spdx.sig \
+            policy-server-${{ matrix.targetarch }}-sbom/policy-server-${{ matrix.targetarch }}-sbom.spdx.json
         env:
           COSIGN_EXPERIMENTAL: 1
 
@@ -112,9 +113,7 @@ jobs:
         with:
           name: policy-server-${{ matrix.targetarch }}-sbom
           path: |
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.json
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.cert
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.sig
+            policy-server-${{ matrix.targetarch }}-sbom
 
   build-container-image:
     name: Build policy server container image
@@ -127,14 +126,9 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: policy-server-x86_64
+      - name: Download artifacts
+        id: download
         uses: actions/download-artifact@v2
-        with:
-          name: policy-server-x86_64
-      - name: policy-server-aarch64
-        uses: actions/download-artifact@v2
-        with:
-          name: policy-server-aarch64
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
       - name: Login to GitHub Container Registry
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image.yml
```diff
@@ -182,3 +176,53 @@ jobs:
...
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.17'
+
+      - name: Install the bom command and prepare destination directory
+        shell: bash
+        run: |
+          set -e
+          go install sigs.k8s.io/bom/cmd/bom@v0.2.2
+          mkdir policy-server-container-image-sbom
```
flavio, line 190: Please, create a dedicated github actions (like we do with kwctl) to install the `sbom` tool. Also, do not build it from source but use a released version. [This](https://github.com/kubernetes-sigs/bom/releases/tag/v0.3.0)  is the latest one.

flavio (commented):
Overall ok. Some extra questions, just to have a better understanding:

* I remember you mentioned other tools that could generate a SBOM for a container image. Why did you pick `sbom`? Note: I have nothing against this choice
* Does the microsoft sbom tool support also container images? I'm asking because it would be nice, one day, to have just one single dependency to create sboms (both for binaries and container images)

jvanz:
>     * I remember you mentioned other tools that could generate a SBOM for a container image. Why did you pick `sbom`? Note: I have nothing against this choice

I've decide to use the `bom` command for Kubernetes because some main reasons:

1.  We are already using it in the controller
2. The `syft` command that I've mentioned does not find any packages in the policy-server container image. I believe this happens because we build a container from the `scratch` image. Which means there are no root fs. Afaics, `syft` checks for the package in the container's root fs. While the Kubernetes `bom` is able to find the files in the container and generate the relationship of that. 
3. It's from Kubernetes community

>     * Does the microsoft sbom tool support also container images? I'm asking because it would be nice, one day, to have just one single dependency to create sboms (both for binaries and container images)

I agree. But we are not using the Microsoft tool anywhere. While we do not get a fix for the [issue](https://github.com/microsoft/sbom-tool/issues/126) in the rust code, I've decided to go with [spdx-sbom-generator](https://github.com/opensbom-generator/spdx-sbom-generator). And this tools does not have SBOM generation for container image. At least, I could not find docs about that. Therefore, in the current state, I need to add other tool in our stack. But I agree that we can migrate to a single tool that attend all our needs in the future.

flavio:
@jvanz thanks for the explanation! Let's go ahead with `sbom` :+1:

jvanz:
This PR now depends on https://github.com/kubewarden/github-actions/pull/39

.github/workflows/container-image.yml
jvanz: Done

.github/workflows/container-image.yml
```diff
@@ -112,9 +113,7 @@ jobs:
         with:
           name: policy-server-${{ matrix.targetarch }}-sbom
           path: |
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.json
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.cert
-            policy-server-${{ matrix.targetarch }}-sbom.spdx.sig
+            policy-server-${{ matrix.targetarch }}-sbom
```
viccuad, line 116: I wonder if this works, or one needs a wildcard pattern to catch the 3 files ending in `.spdx.{json,cert,sig}`.

.github/workflows/container-image.yml
```diff
@@ -90,13 +90,14 @@ jobs:
           make download-spdx-sbom-generator sbom
```
viccuad, line 90: I think this now could be substituted by the new github action, [bom-generator-installer](https://github.com/kubewarden/github-actions/blob/0e8fe58ef0bdcba71e355dfa4bb33c5ae54e7ec4/sbom-generator-installer/action.yaml).

.github/workflows/container-image.yml
viccuad: nevermind, this is a directory, and the files are inside of it.

.github/workflows/container-image.yml
```diff
@@ -182,3 +176,49 @@ jobs:
...
+
+      - name: Create SBOM file for the container  image
+        if: ${{ startsWith(github.ref, 'refs/heads/') }}
+        shell: bash
+        run: |
+          set -e
+          bom generate -n https://kubewarden.io/ --image ghcr.io/${{ github.repository_owner }}/policy-server@${{ steps.build-latest.outputs.digest }} -o policy-server-container-image-sbom/policy-server-container-image-sbom.spdx
+
+      - name: Create SBOM file for the container  image
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
```
viccuad, line 196: I think this step and the previous one can be collapsed into one, and no `if ${{ startsWith(github.ref ... ` is needed.
This is because the workflow is run for `on.push.branches["main"]` and `on.push.tags[ "v*" ]` which I think is equivalent.

viccuad (approved):
Approving, I'm happy with getting this merged as it is. We can do changes in a followup PR if needed.

.github/workflows/container-image.yml
flavio: Yes, I think you are right. We could have one single "Create SBOM file for the container image" step, without any `if` statement.

Is that what you mean?

viccuad: yep, correct.

flavio:
@viccuad can you take another look? I did the changes we proposed

viccuad (approved):

viccuad:
Thanks, merging!

Issue #254, opened 2022-05-17
Add timeout to policy evaluation


jvanz:
To help the mitigation of the [threat #2](https://github.com/kubewarden/rfc/blob/main/rfc/0006-threat-model.md#threat-2---attacker-passes-workloads-which-require-complex--processing-causing-timeouts) of the threat model, the policy server should allow the operator define a timeout for the policy evaluation. If a policy takes more then the defined timeout, the request is rejected. 

**NOTE:** This is an issue created from RFC discussing the admission control threat model. It's created to allow the Kubewarden team discuss the proposed mitigation further and start to work on each actionable item when possible

## Actions

* [x] Update policy-evaluator: https://github.com/kubewarden/policy-evaluator/pull/224
* [x] Update policy server: https://github.com/kubewarden/policy-server/pull/389
* [x] Update docs: https://github.com/kubewarden/docs/issues/157
* ~~ [ ] Blog post: https://github.com/kubewarden/kubewarden.io/issues/165 ~~

flavio:
I think there are basically two options:

* Implement a timer
* Try to use something fancy like the [fuel](https://docs.rs/wasmtime/0.39.1/wasmtime/struct.Config.html#method.consume_fuel) concept built into wasmtime

flavio:
I wonder if "timeout protection" is a good name for this feature :thinking: 

I think we should think about another one, because this is a bit confusing

viccuad:
I'm having trouble finding a better wording for "if policy timeout reached, reject request".
Ideas: `rejection by timeout`, `timeout exclusion`.
I still prefer `timeout protection`.

viccuad:
https://github.com/kubewarden/policy-server/pull/389 got prematurely merged, https://github.com/kubewarden/policy-server/pull/396 reverted it.

Re-opening card again. We would need to create a new PR analogous to https://github.com/kubewarden/policy-server/pull/389.

flavio:
reopening, there are other tasks that have to be done

PR #329, opened 2022-09-23
Change architecture of Workers pool

flavio:
This is a big refactoring of the WorkerPool and the Worker structs.

The major changes done to WorkerPool:

* Perform the validation of policies settings only once per policy.    Prior to that, each worker validated the settings of all the policies.
* Precompile all the WebAssembly modules once, then give the precompiled      modules to the workers. This reduces the bootstrap time **a lot**.
    
About the latter point, creating a `wasmtime::Module` by reading the     contents of the `.wasm` file is an expensive operation. Prior to that,    the same WebAssembly module was compiled by each worker.
    
Note, we cannot rely on Wasmtime `cache` feature because, given the number of workers we start, there would be quite some resource contention among them.
    
Changes done to the Worker:
    
* Create one single `wasmtime::Engine` to be shared among all the Modules hosted by the Worker. This reduces resource usage and is needed to implement the "limit execution time of the policies"
feature.
* Instantiate the `wasmtime::Module` instances by using the precompiled objects prepared by the WorkerPool. This saves an insane amount of time at bootstrap time
* Do not perform settings validation anymore. They have already been validated prior to the creation of the Worker

Finally, this is required to allow us to implement https://github.com/kubewarden/policy-server/issues/254


flavio:
This depends on https://github.com/kubewarden/policy-evaluator/pull/186

viccuad (approved):
LGTM! Nice PR, as the other!

PR #350, opened 2022-10-24
fix: create SBOM for tagged container image

flavio:
Ensure the SBOM for the tagged container image (like `policy-server:2.0.0`) is created.

Prior to this commit, the creation of the SBOM failed because the signing action was trying to obain the digest of the `latest` container image.

This fixes build errors like this one: https://github.com/kubewarden/policy-server/actions/runs/3298307317/jobs/5440643528


raulcabello (approved):
LGTM

PR #356, opened 2022-10-26
fix: add binaries signatures to the same zip file

flavio:
Add the signagures of the `policy-server` binaries to the same zip file. Previously, they where put into a dedicated zip file.

This change is needed to make CLOMonitor that we are signing all our assets.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -86,37 +86,30 @@ jobs:
           RUSTFLAGS: "-C link_arg=-lgcc -C link_arg=-specs -C link_arg=/usr/lib/${{ matrix.targetarch}}-linux-musl/musl-gcc.specs ${{ matrix.rustflags }}"
         run: |
           cargo build --release --target ${{ matrix.targetarch }}-unknown-linux-musl
-          mv target/${{ matrix.targetarch }}-unknown-linux-musl/release/policy-server policy-server-${{ matrix.targetarch }}
+          mkdir policy-server-${{ matrix.targetarch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/container-image.yml
```diff
@@ -86,38 +86,31 @@ jobs:
...
+          mv target/${{ matrix.targetarch }}-unknown-linux-musl/release/policy-server policy-server-${{ matrix.targetarch }}/
 
       - name: Generate SBOM
         run: |
           spdx-sbom-generator -f json
           # SBOM files should have "sbom" in the name due the CLO monitor
           # https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom
-          mkdir policy-server-${{ matrix.targetarch }}-sbom
-          mv bom-cargo.json policy-server-${{ matrix.targetarch }}-sbom/policy-server-${{ matrix.targetarch }}-sbom.spdx.json
+          mv bom-cargo.json policy-server-${{ matrix.targetarch }}/policy-server.spdx.json
```
viccuad, line 97: doesn't this overwrite `policy-server.spdx.json` for the different targetarchs?

flavio: This isn't happening as far as I've seen.

We create a directory like `policy-server-x86_64`, inside of this directory we put the binary for that platform, the SBOM of that binary and all the files produced by `cosign` when signing the SBOM.

Later we create a dedicated zip file with the SBOM files (and their signatures) for the container image.

Does that help?

viccuad: doh, indeed!

viccuad (approved):

.github/workflows/container-image.yml
flavio: Solved as reported on slack

PR #357, opened 2022-10-26
v1.3.0-rc3 release

flavio:
Prepare 1.3.0-rc3


viccuad (approved):

viccuad (changes requested):
I think we are missing the version bump in Cargo.toml

viccuad (approved):

PR #384, opened 2022-12-13
feat: audit endpoint

flavio:
Add new rest API endpoint called `/audit`, similar to the current `/validate` one.

This endpoint receives POST requests similar to the ones generated by the Kubernetes API server. All the policies hosted by the Policy Server are exposed via `/audit/POL_ID` endpoints.

Policies in monitor mode behave as if they were in protect mode, to produce rejections.

The endpoint generates its own Prometheus metrics and OpenTelemetry traces to not be mixed with `/validate` ones.

While working on that, I cleaned up the code based and introduced new unit tests.

Fixes https://github.com/kubewarden/kubewarden-controller/issues/337


jvanz (approved):
LGTM! :D

viccuad (approved):
LGTM!

PR #389, opened 2022-12-19
feat: policy timeout protection

flavio:
Ensure a policy cannot run more than a given amount of time.

This is required to fix https://github.com/kubewarden/policy-server/issues/254

Note 1: this PR builds on top of https://github.com/kubewarden/policy-server/pull/388 to ensure all the tests are passing

Note 2: this PR depends on https://github.com/kubewarden/policy-evaluator/pull/224

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,12 +25,12 @@ test: fmt lint
 
 .PHONY: e2e-tests
 e2e-tests: target/release/policy-server
-	pushd ./e2e-tests && make test
+	make -C ./e2e-tests test
 
 .PHONY: clean
 clean:
 	cargo clean
-	pushd e2e-tests && make clean
+	make -C e2e-tests clean
 
 .PHONY: tag
 tag:
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,15 @@
+SHELL := /bin/bash
 IMG ?= policy-server:latest
 BINDIR ?= bin
 SBOM_GENERATOR_TOOL_VERSION ?= v0.0.15
 
-.PHONY: build
-build:
+SOURCE_FILES := $(shell test -e src/ && find src -type f)
+
+target/release/policy-server: $(SOURCE_FILES) Cargo.*
 	cargo build --release
+
+.PHONY: build
+build: target/release/policy-server
 
 .PHONY: fmt
 fmt:
@@ -18,9 +23,14 @@ lint:
 test: fmt lint
 	cargo test --workspace
 
+.PHONY: e2e-tests
+e2e-tests: target/release/policy-server
+	pushd ./e2e-tests && make test
+
 .PHONY: clean
 clean:
 	cargo clean
+	pushd e2e-tests && make clean
 
 .PHONY: tag
 tag:
--- a/e2e-tests/01-setup.yml
+++ b/e2e-tests/01-setup.yml
@@ -1,0 +1,36 @@
+name: Policy Server end-to-end setup
+vars:
+  pidfile: policy-server.pid
+
+testcases:
+- name: Build policy-server binary with release mode. Can take quite some time...
+  steps:
+  - script: cargo build --release
+    type: exec
+
+- name: Start policy-server
+  steps:
+  - script: |
+      cargo run --release -- \
+        --policies test_data/policies.yaml \
+        --policies-download-dir `pwd`/e2e-tests-policies \
+        --policy-timeout 2 \
+        --workers 1 \
+        --daemon true \
+        --daemon-stdout-file policy-server.out \
+        --daemon-stderr-file policy-server.err \
+        --daemon-pid-file {{ .pidfile }}
+    type: exec
+
+- name: Wait for policy-server to be ready
+  steps:
+  - type: http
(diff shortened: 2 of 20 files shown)


src/cli.rs
```diff
@@ -180,6 +180,21 @@ pub(crate) fn build_cli() -> Command {
...
+                .required(false)
+                .help("Disable policy timeout protection"),
+        )
+        .arg(
+            Arg::new("policy-timeout")
+                .long("policy-timeout")
+                .env("KUBEWARDEN_POLICY_TIMEOUT")
+                .value_name("MAXIMUM_EXECUTION_TIME_SECONDS")
+                .default_value("2")
+                .help("Interrupt policy evaluation after the given time"),
```
jvanz, line 196: What's the time unit? I guess is in seconds. Am I right?

```suggestion
                .help("Interrupt policy evaluation after the given time (in seconds)"),
```

jvanz (approved):

flavio:
@jvanz please, take a look at the latest commits I've pushed. They add the e2e tests

@kravciak this could be interesting for you as well, the commit introducing the e2e tests is https://github.com/kubewarden/policy-server/pull/389/commits/955ac59f6bad648b38e2acc97e90ce623452367f

Makefile
```diff
@@ -18,9 +23,14 @@ lint:
 test: fmt lint
 	cargo test --workspace
 
+.PHONY: e2e-tests
+e2e-tests: target/release/policy-server
+	pushd ./e2e-tests && make test
```
jvanz, line 28:
```suggestion
	make -C ./e2e-tests test
```

Makefile
```diff
@@ -18,9 +23,14 @@ lint:
...
+
 .PHONY: clean
 clean:
 	cargo clean
+	pushd e2e-tests && make clean
```
jvanz, line 33:
```suggestion
	make -C e2e-tests clean
```

e2e-tests/01-setup.yml
```diff
@@ -0,0 +1,36 @@
+name: Policy Server end-to-end setup
+vars:
+  pidfile: policy-server.pid
+
+testcases:
+- name: Build policy-server binary with release mode. Can take quite some time...
+  steps:
+  - script: cargo build --release
+    type: exec
```
jvanz, line 9: Why do we need to build it here if the makefile already does that? Is the makefile dependency just to check if the code is building?

src/main.rs
```diff
@@ -64,6 +64,50 @@ fn main() -> Result<()> {
...
+                    "'policy-timeout' value cannot be converted to unsigned int: {}",
+                    e
+                ));
+                unreachable!()
+            }
+        }
+    };
+
+    // Run in daemon mode if specified by the user
+    if matches.contains_id("daemon") {
```
jvanz, line 87: Is the daemonization of the process required just because of the tests? Or is there another reason?

jvanz (changes requested):

e2e-tests/01-setup.yml
flavio: Building in release mode takes a long time. I prefer to have a dedicated step inside of the GH action because that provides a better feedback.

src/main.rs
flavio: Yes, that's only needed by the e2e tests (so far). This made super easy to start policy-server from venom

flavio:
@jvanz I've applied all the changes requested, plus I've managed to fix the e2e-tests :partying_face:

src/worker_pool.rs
```diff
@@ -217,6 +256,21 @@ impl WorkerPool {
         // We can start waiting for admission review requests to be evaluated
         let mut next_worker_id = 0;
 
+        if self.policy_evaluation_limit_seconds.is_some() {
```
viccuad, line 259: shouldn't this be before doing policy init, and before `verify_policy_settings()` that would already check against an epoch?

viccuad (approved):
LGTM, it's super nice to have e2e tests now :tada: !

(missing consuming correct policy-evaluator repo)

jvanz (approved):

viccuad:
Moving the PR back to review, so @flavio can have a look at the comments. For merging, it will need a new PR.

jvanz:
> This PR got prematurely merged, #396 reverted it.
> 
> Moving the PR back to review, so @flavio can have a look at the comments. For merging, it will need a new PR.

Sorry @viccuad your "LGTM" made me think that the question was not something blocking the merge of the changes.

src/worker_pool.rs
flavio: Good catch, the code you reviewed did NOT have timeouts applied while doing the validation of the policy settings. Unfortunately we cannot move this block of code earlier, because the validation of the settings is done by a single engine. This thread instead goes over all the engines of the workers and increases their internal timer.

I'm going to make a dedicated commit that addresses this issue

flavio: fixed with 06ba04148125211ce24a0a9d69c3b1b05c566fa7

Issue #376, opened 2022-11-23
Trigger chart release when policy-server is released


jvanz:
Updates the release CI/CD pipeline to trigger the `kubewarden-controller` and `kubewarden-crds` [charts](https://github.com/kubewarden/helm-charts) updates.

We might use this [action](https://github.com/peter-evans/repository-dispatch ) to dispatch workflows

jvanz:
Moving this to block until we fix this [issue](https://github.com/kubewarden/policy-server/pull/392#issuecomment-1373964388)

PR #392, opened 2022-12-23
Trigger Helm chart release.

jvanz:
## Description

Updates the release CI job to trigger an event to the Helm chart repository to ensure that the major and minor releases will be in sync between all Kubewarden components.

Fix https://github.com/kubewarden/policy-server/issues/376

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,6 +41,22 @@ jobs:
       - name: 'Unzip build artifacts'
         run: unzip -n "policy-server-*.zip" -d artifacts
 
+      - name: Get old release tag
+        id: get_old_release_tag
+        uses: actions/github-script@v6
+        with:
+          script: |
+            let release = await github.rest.repos.getLatestRelease({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+            });
+
+            if (release.status  === 200 ) {
+              core.setOutput('release_tag', release.data.tag_name)
+              return
+            }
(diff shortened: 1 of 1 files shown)


viccuad (approved):
LGTM!

Analogous to the one in kubewarden-controller repo. Note though that the one from controller repo failed because of the secret.

jvanz:
> Analogous to the one in kubewarden-controller repo. Note though that the one from controller repo failed because of the secret.

That's weird. We are using the same token used by the bot that triggers jobs to update the policy hub repository. @flavio we need to review the token permissions.

jvanz:
Moving this to block until we fix the secret permissions.

flavio:
It looks like the PAT has the right privileged enabled:

image

flavio:
I think I figured out why this is happening: the token belongs to @chimera-kube-bot. Hence the bot must have write permissions to some of the repository where the target workflow belongs to. In this case, I guess we must give it write access to the https://github.com/kubewarden/helm-charts repository.
Am I right, is there some other repo that we should grant write access to?

jvanz:
> Am I right, is there some other repo that we should grant write access to?

On top of my head, I cannot think in any other one. As the request will be trigger from the components to the helm-chart repository. I think it only need write access to the helm-chart repository for now.

flavio:
@jvanz I've just given write access to @chimera-kube-bot  to the `hem-charts` repository. Let me know if that fixes the issue :crossed_fingers:

PR #407, opened 2023-01-25
chore: ignore twoway warning

flavio:
Cargo audit is now failing because of [RUSTSEC-2021-0146](https://rustsec.org/advisories/RUSTSEC-2021-0146).

This is about `twoway` not being maintained anymore. This is a transitive dependency of `multipart`, which is a dependency of `warp`.

[This is the GH issue](https://rustsec.org/advisories/RUSTSEC-2021-0146) that describes the problem.

There's nothing we can do in the meantime. We have to ignore this warning otherwise it will not be possible to merge PRs.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -7,8 +7,6 @@ ignore = [
                        # to API breakages.
                        #
                        # This is a transitive depependency
-  "RUSTSEC-2021-0139", # This is about `ansi_term` not being maintained anymore
-                       # This is a transitive depependency
   "RUSTSEC-2021-0146", # This is about `twoway` not being maintained anymore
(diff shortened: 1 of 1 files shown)


.cargo/audit.toml
```diff
@@ -9,4 +9,8 @@ ignore = [
                        # This is a transitive depependency
   "RUSTSEC-2021-0139", # This is about `ansi_term` not being maintained anymore
```
viccuad, line 10: I think we can drop this, cargo.lock doesn't contain `ansi_term` anymore, and now has `nu_ansi_term`.

flavio: Good catch. Yes, we can drop that

Issue #413, opened 2023-02-09
Rego policies cannot be instantiated when timeout protection is enabled


flavio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Since release 1.5, Policy Server has the timeout protection feature enabled. Unfortunately, when this feature is enabled the instantiation of Rego-based policies failes with this error:

```console
2023-02-08T18:53:28.610850Z INFO policy_server::policy_downloader: policies download download_dir="/tmp" policies_count=1 status="init"
2023-02-08T18:53:28.708822Z INFO policy_server::policy_downloader: policy download name="clusterwide-disable-lbs" path="/tmp/registry/ghcr.io/kubewarden/policies/disallow-service-loadbalancer:v0.1.5" sha256sum="1196136c8fe90895af5c9569077008f68db081db5b4027db735f89acdb3aae47" mutating=false
2023-02-08T18:53:28.709002Z INFO policy_server::policy_downloader: policies download status="done"
2023-02-08T18:53:28.709102Z INFO policy_server: kubernetes poller bootstrap status="init"
2023-02-08T18:53:28.709285Z INFO policy_server::kube_poller: spawning cluster context refresh loop
2023-02-08T18:53:28.709298Z INFO policy_server: kubernetes poller bootstrap status="done"
2023-02-08T18:53:28.709699Z INFO policy_server: CallbackHandler task status="init"
2023-02-08T18:53:28.709802Z INFO policy_server: worker pool bootstrap status="init"
2023-02-08T18:53:29.050928Z ERROR policy_server::worker_pool: cannot validate policy settings error=[clusterwide-disable-lbs] cannot create PolicyEvaluator: wasm engine error: linker cannot create instance: error while executing at wasm backtrace:
0: 0xc812 - <unknown>!_initialize
2023-02-08T18:53:29.051180Z ERROR policy_server: [clusterwide-disable-lbs] cannot create PolicyEvaluator: wasm engine error: linker cannot create instance: error while executing at wasm backtrace:
0: 0xc812 - <unknown>!_initialize
```

### Expected Behavior

Rego based policies should work just fine, also when the timeout protection is enabled

### Steps To Reproduce

Start Policy Server with a policies.yml file like the following one:

```yaml
loadbalancer:
  url: ghcr.io/kubewarden/policies/disallow-service-loadbalancer:v0.1.5
```
From the policy-server repo:
```console
 cargo run --release -- --policies bug-policies.yml --log-level debug
[...]

PR #414, opened 2023-02-09
test: extend e2e tests

flavio:
Ensure Rego policies are being tested as part of the e2e tests.

This would have allowed us to spot https://github.com/kubewarden/policy-server/issues/413 immediately.

**WARNING:** the e2e tests inside of this PR will keep failing until the fix for https://github.com/kubewarden/policy-server/issues/413 lands into this branch. That's a feature! :smile:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,9 +69,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.68"
+version = "1.0.69"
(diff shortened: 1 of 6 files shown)


flavio:
@jvanz I've just updated the `policy-evaluator` dependency to be consumed. This is going to fix the e2e tests. Once this gets merged I'll bump the policy-server version and tag a new patch release

PR #415, opened 2023-02-10
fix: make e2e tests more scrupulous

flavio:
Extend the tests to look also at the reason that lead a request to be rejected.

This would have allowed us to spot https://github.com/kubewarden/policy-evaluator/issues/239
ahead of time.

**Note:** the e2e tests of this PR will keep failing until the update policy-evaluator to include the fix of the bug linked above. That's a feature :smile:


jvanz (approved):

flavio:
I've updated the policy-evaluator dependency, plus bumped the patch release of policy-evalutor

I'll wait for all the tests to be green, then I'll merge this PR and tag a new release

PR #447, opened 2023-04-05
Update DEVELOPMENT.md

viccuad:
## Description

Partial docs for the debug of pods.

## Test


jvanz (approved):
Thanks for documenting that! This is really important. 

I have just a question... as far as I can remember, this commands were failing in @flavio's demonstration. Was that a Flavio's environment issues?

flavio:
I found what I was wrong, it was all caused by the group I assigned the `kw` to.

I've pushed a commit with all the fixes (there were too many to be added via suggestion) and some integration.

I've also managed to get `strace` to work

flavio (approved):
Approving, but I'm biased :)

PR #461, opened 2023-05-12
ci: Build also `:latest-feat-foo` images 

viccuad:
## Description

ci: Set image:latest tag depending on branch name, and run the workflow also on `feat-` branches.
This gives us `:latest-feat-foo` image tags.

## Test

Tested on kubewarden-controller repo.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
     - main
+    - "feat-**"
     tags:
     - 'v*'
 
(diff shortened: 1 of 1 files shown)


viccuad:
I'm happy closing this PR without merging if we don't see the need for it.

jvanz:
> I'm happy closing this PR without merging if we don't see the need for it.

I like this change. I use features branches and container image for them to test my changes. Thus, adding this in CI is welcome to me. :)

jvanz (approved):

flavio (approved):
I like it too

PR #473, opened 2023-06-19
wasi policies

flavio:
Support policies that implement the new WASI protocol.

This policy execution mode is not the best one, but it allows the execution of WASM modules that do not leverage waPC. This is the case of modules built by the official Go compile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -413,7 +413,7 @@ checksum = "a3e2c3daef883ecc1b5d58c15adae93470a91d425f3532ba1695849656af3fc1"
 [[package]]
 name = "burrego"
 version = "0.3.4"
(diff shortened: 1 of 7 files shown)


flavio:
Converted to draft. The e2e tests will fail because it's missing the WASI policy to be used. I need to create a WASI policy and host it on ghcr, however I cannot do that because there's no released version of kwctl that can annotate a policy (see https://github.com/kubewarden/kwctl/pull/518)

jvanz:
It looks fine. Let's wait for the kwctl PR get merge and then we can run it on CI and other machines.

flavio:
Moving back to pending review, I've pushed a test policy that is now used by the e2e tests. IMHO we can merge this PR if everything is green

jvanz:
Everything is green and I can build and run test locally as well. Merging this now...

PR #496, opened 2023-07-17
fix(perf): reduce memory usage

flavio:
Reduce the amount of memory being used by policy-server at runtime.

This table contains a comparison of the memory usage (Mb) before and after this optimization:

| Workers (#) | Vanilla (Mb) | Optimized (Mb) | Improvement (%) |
|-------------|--------------|----------------|-----------------|
| 1           | 634.356      | 506.484        | -20.16          |
| 2           | 790.672      | 598.902        | -24.25          |
| 3           | 954.726      | 680.124        | -28.76          |
| 4           | 1108.042     | 774.498        | -30.10          |

comparison

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1552,6 +1552,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cda653ca797810c02f7ca4b804b40b8b95ae046eb989d356bce17919a8c25499"
 
 [[package]]
+name = "flate2"
+version = "1.0.26"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b9429470923de8e8cbd4d2dc513535400b4b3fef0319fb5c4e1f520a7bef743"
+dependencies = [
+ "crc32fast",
+ "miniz_oxide",
+]
+
+[[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jvanz (approved):
That's a nice fix! I'm glad that you have done it.

However, it still looks a lot of memory consumption. How many policies we have in the test from the chart?

flavio:
Yes, this not the solution to the problem, just a quick fix.

Sorry, in the rush of pushing things I forgot to add more details:

These are the policies I loaded:

```yaml
pod-privileged-protect-mode:
  url: ghcr.io/kubewarden/tests/pod-privileged:v0.2.1
  policyMode: protect
  allowedToMutate: false
  settings: {}

pod-privileged-protect-mode2:
  url: ghcr.io/kubewarden/tests/pod-privileged:v0.2.1
  policyMode: protect
  allowedToMutate: false
  settings: {}

pod-privileged-protect-mode3:
  url: ghcr.io/kubewarden/tests/pod-privileged:v0.2.1
  policyMode: protect
  allowedToMutate: false
  settings: {}

pod-privileged-protect-mode4:
  url: ghcr.io/kubewarden/tests/pod-privileged:v0.2.1
  policyMode: protect
  allowedToMutate: false
  settings: {}

pod-privileged-monitor-mode:
  url: ghcr.io/kubewarden/tests/pod-privileged:v0.2.1
  policyMode: monitor
  allowedToMutate: false
  settings: {}

sleep:
  url: ghcr.io/kubewarden/tests/sleeping-policy:v0.1.0
  policyMode: protect
  allowedToMutate: false
  settings:
    sleepMilliseconds: 2000

disallow-service-loadbalancer:
  url: ghcr.io/kubewarden/tests/disallow-service-loadbalancer:v0.1.5
  policyMode: protect
  allowedToMutate: false

flux:
  url: ghcr.io/kubewarden/tests/go-wasi-template:v0.1.0
  policyMode: protect
  allowedToMutate: true
  settings:
    requiredAnnotations:
      "fluxcd.io/cat": "felix"
```

The values reported are an average of 5 measurements. I also calculated standard deviation, which is also shown as an error bar on the Y axis.

More work can be done to improve the memory usage, but this would be more invasive. Something that needs better planning.

flavio:
**UPDATE/DISCLAIMER:** I looked back at the performance improvements of this PR because of some recent investigation work I've done. The high memory usage reported is caused by the `go-wasi-template` policy. This is a WASI policy produced by the native Go compiler, which is pretty big (~20 Mb).

Using the policies from https://github.com/kubewarden/policy-server/pull/496#issuecomment-1638122503 leads to a memory usage of ~491 Mb, just with a single worker.
Removing the WASI policy leads to a memory usage of ~124 Mb, always with a single worker.

Go WASI policies are tricky, we should not consider them the norm when we do memory usage analysis

PR #497, opened 2023-07-18
refactor: new container image build process

flavio:
Prior to this PR, building the container image was done performing a complex cross-compilation build of the policy-server binary.
That required a special environment to be setup inside of the GitHub action. This environment was a bit scary to maintain, it required a specific set of repositories and workarounds in place, all of them done inside of our GitHub action.

Because of that, performing a local build of the container image was not doable.

This change allows to build the container image doing a simple `docker build`. Moreover, the complex build environment is no longer needed.

This PR changes the GitHub actions used to build and release the container image. I've taken the ones used by kubewarden-controller and replaced all the occurrences of `kubewarden-controller` with `policy-server`.


.github/workflows/container-build.yml
```diff
@@ -0,0 +1,51 @@
...
+        description: "Container image digest"
+        value: ${{jobs.build.outputs.digest}}
+  push:
+    branches:
+      - "*"
+
+jobs:
+  build:
+    name: Build
+    uses: kubewarden/policy-server/.github/workflows/container-image.yml@main
```
flavio, line 16: This is taken from the kubewarden-controller, I don't know if we should force the usage of `@main` here (or in other places where we reuse a local (as in from the same repository) workflow.

I think we should instead use the `./.github/workflows/{filename}` syntax, because that ensures the workflow is taken from the same git rev of the "main" workflow.

See [this section](https://docs.github.com/en/actions/using-workflows/reusing-workflows) of the official docs:

> If you use the second syntax option (without `{owner}/{repo}` and `@{ref}`) the called workflow is from the same commit as the caller workflow.

flavio:
JFYI: I've built the container image locally, using `docker build`, pushed it to a registry and then consumed it from a Kubernetes cluster. Everything seemed to be working fine

.github/workflows/container-build.yml
```diff
@@ -0,0 +1,51 @@
...
+
+on:
+  workflow_call:
+    outputs:
+      digest:
+        description: "Container image digest"
+        value: ${{jobs.build.outputs.digest}}
+  push:
+    branches:
+      - "*"
```
viccuad, line 11: I think we should limit this to `main`. If not, the `Retrieve tag name` step of the other workflow, the container-image.yml one, will overwrite the `:latest` tag with whatever build from whatever branch that has run the workflow the latest.

.github/workflows/release.yml
```diff
@@ -105,20 +121,20 @@ jobs:
...
 
       - name: Trigger chart update
-        uses: peter-evans/repository-dispatch@26b39ed245ab8f31526069329e112ab2fb224588 # v2.1.1
+        uses: peter-evans/repository-dispatch@4c03f6fcc3d786d9feef89ed5544ccf37d52e232
         with:
           token: ${{ secrets.WORKFLOW_PAT }}
           repository: "${{github.repository_owner}}/helm-charts"
           event-type: update-chart
-          client-payload: '{"version": "${{ github.event.workflow_run.head_branch }}", "oldVersion": "${{ steps.get_old_release_tag.outputs.release_tag }}", "repository": "${{ github.repository }}"}'
+          client-payload: '{"version": "${{ github.ref_name }}", "oldVersion": "${{ steps.get_last_release_tag.outputs.old_release_tag }}", "repository": "${{ github.repository }}", "crds_asset_id": "${{steps.upload_release_assets.outputs.crds_asset_id}}"}'
```
viccuad, line 140: it seems to me that this is a left-over from the controller one; we don't have `crds_asset_id` here for example. I'm not sure about the `github.ref_name` change, though.

.github/workflows/container-build.yml
jvanz: That's true. Maybe we can do something similar to what we have in the controller. Thus, we can have a container image for each feature branch. Which is nice for testing and validation of our changes.

jvanz (changes requested):
Thanks for improving this! :)

.github/workflows/container-build.yml
flavio: this should be solved now

flavio:
I made a mistake when I copied the GH actions from the kubewarden-controller, I used the files of the v1.6.0 release instead of the latest ones from the main branch.

I've done the sync again, please take a look at the new files.

I'll squash the commits before doing the final merge, the goal is to have 2 commits: one for the dockerfile and the other for the GH actions

flavio:
> There's currently a build issue with ARM64. I'm working on it

This should have been fixed, at least it was working on a arm64 host I used.

.github/workflows/container-build.yml
flavio: I've updated all the `uses` references inside of the GH actions to make use of the local file

.github/workflows/release.yml
```diff
@@ -105,20 +121,20 @@ jobs:
...
             const {RELEASE_ID} = process.env
+            const {TAG_NAME} = process.env
             github.rest.repos.updateRelease({
               owner: context.repo.owner,
               repo: context.repo.repo,
               release_id: `${RELEASE_ID}`,
               draft: false,
-              tag_name: `${{ github.event.workflow_run.head_branch }}`,
-              name: `${{ github.event.workflow_run.head_branch }}`,
-              prerelease: `${{ contains(github.event.workflow_run.head_branch, '-alpha') || contains(github.event.workflow_run.head_branch, '-beta') || contains(github.event.workflow_run.head_branch, '-rc') }}`,
```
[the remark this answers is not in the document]
flavio: oh, the kubewarden-controller doesn't have this flag, should we add it?

https://github.com/kubewarden/kubewarden-controller/blob/a0ad85cffd6b100e2d0955082ad8ea2c9a919b8b/.github/workflows/release.yml#L128-L141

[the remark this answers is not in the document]
viccuad: I would leave it. This only changes if the GH release is marked as pre-release or not. Also, I don't think we are using that metadata in our own automation.

.github/workflows/container-build.yml
viccuad: Agree with this change. Don't see any security issue, and at least it makes it easier to work with workflows over forks.

Dockerfile
```diff
@@ -1,21 +1,27 @@
...
+RUN rustup target add $(arch)-unknown-linux-musl
 
-# amd64-specific
-FROM scratch AS build-amd64
-COPY --from=common /etc/passwd /etc/passwd
-COPY --from=common /etc/group /etc/group
-COPY --chmod=0755 policy-server-x86_64 /policy-server
+# Fix ring building using musl - see https://github.com/briansmith/ring/issues/1414#issuecomment-1055177218
+RUN apt-get update && apt-get install musl-tools clang llvm -y
+ENV CC="clang"
```
viccuad, line 9: I suppose this is the only way of setting the compiler, and we can't do it on the cargo.toml file.
It would be nice to have the same way of building a local release binary, and the container image (apart from having DRY).

.github/workflows/release.yml
```diff
@@ -49,47 +95,17 @@ jobs:
...
         id: upload_release_assets
         uses: actions/github-script@v6
         with:
           script: |
             let fs = require('fs');
-            // The artifact directory is defined in the previous step. When the
-            // zip files are expanded.
-            let files = fs.readdirSync("artifacts")
-            // Envvar exported in the step para the draft release is found
+            let files = ['policy-server-sbom.spdx', 'policy-server-sbom.spdx.cert', 'policy-server-sbom.spdx.sig']
```
viccuad, line 104: I don't think this are the only files, on the release page assets I see there are different ones depending on the arch, plus we have distinct sbom files for the binary and the container image.

I think we should keep them.

I'm suspecting that this informed the previous design; we needed to build the binary out of the Dockerfile, so we could easily gather its SBOMs, per arch.

Then, we built the container image, and gathered its SBOMs, too. Here we were just copying the binary.

Then for the release, we would upload every file in `artifacts/`.


So, 
Is there need for separate SBOMs of the binary (and build process), and the build inside the container (that happens in this PR)?
If we build the bin inside the container stage, that later we discard, do we get the correct SBOMs?

viccuad (changes requested):

.github/workflows/release.yml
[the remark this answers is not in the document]
jvanz: > oh, the kubewarden-controller doesn't have this flag, should we add it?

Yes, we should. This explains why the[ `v1.7.0-rc1` ](https://github.com/kubewarden/kubewarden-controller/releases/tag/v1.7.0-rc1)is set as `latest`. :facepalm:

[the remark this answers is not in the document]
jvanz: I've fixed the releases manually and open PRs fixing the CI in the controller and kwctl repositories.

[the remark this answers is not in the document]
jvanz: https://github.com/kubewarden/kubewarden-controller/pull/486
https://github.com/kubewarden/kwctl/pull/543

.github/workflows/release.yml
```diff
@@ -1,23 +1,69 @@
...
+            --image "ghcr.io/${{github.repository_owner}}/policy-server@${{ needs.container-build.outputs.digest }}" \
+            . > policy-server-sbom.spdx
+
+      - name: Sign BOM file
+        run: |
+          cosign sign-blob --yes --output-certificate policy-server-sbom.spdx.cert \
+            --output-signature policy-server-sbom.spdx.sig \
+            policy-server-sbom.spdx
+        env:
+          COSIGN_EXPERIMENTAL: 1
```
jvanz, line 63:
```suggestion
```

We do not need this. Keyless signing is not experimental anymore. 

https://blog.sigstore.dev/cosign-2-0-released/

jvanz (changes requested):

.github/workflows/release.yml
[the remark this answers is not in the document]
flavio: cool, marking this comment as resolved

Dockerfile
flavio: Unfortunately this cannot be set into the `.cargo/config.toml`. I really tried hard, but I couldn't find a way to do that.

The local build (`cargo build`) is going to use gcc and is not going to use musl. Do we want to change that to use musl too? I think this should be possible, but I'm not entirely sure about that.

However, keep in mind we will introduce the following prerequisites:

* Have clang installed
* Have the musl target added
* Have something like [direnv](https://direnv.net/) installed, so that we can ensure the `CC` is set accordingly

Do you want me to give it a try?

.github/workflows/release.yml
flavio: Ok, we have to find a way to generate the SBOMs for the binaries. We can't go back to the old build system and keep this change, **unless** we want to end up with something like:
* current `main` branch
* add `Dockerfile.next` which is the `Dockerfile` of this PR and add minimal amount of files needed by it (like the `.cargo/config.toml` - which could also be dropped if we add more `ENV` directives inside of the `Dockerfile.next`)

Let me research a bit how we can approach this problem, without having to go back to the scenario I just illustrated

flavio:
I've just propagated the [changes](https://github.com/kubewarden/kubewarden-controller/pull/487) done to the kubewarden-controller GH actions here.

I've done a quick build against main, with only amd64 enabled. Everything worked fine. Now I'm running a build (always against main) for both amd64 and arm64. I've also tagged a fake 1.7.1 on my fork to test the release process.

In the meantime you can review the PR

.github/workflows/container-image.yml
```diff
@@ -1,262 +1,117 @@
...
           context: .
-          load: true
+          file: ./Dockerfile
+          platforms: linux/amd64, linux/arm64
+          push: true
           tags: |
-            ghcr.io/${{ github.repository_owner }}/policy-server:${{ env.TAG_NAME }}
-      - name: Test container image
-        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+            ghcr.io/${{github.repository_owner}}/policy-server:${{ env.TAG_NAME }}
```
viccuad, line 82: Just a realization:

```suggestion
            ghcr.io/${{github.repository_owner}}/{{ github.event.repository.name }}:${{ env.TAG_NAME }}
```

(and elsewhere)
And at this point, the workflows are the same between repos. I don't want to do that right now though.

.github/workflows/release.yml
```diff
@@ -105,20 +110,21 @@ jobs:
...
 
       - name: Trigger chart update
-        uses: peter-evans/repository-dispatch@26b39ed245ab8f31526069329e112ab2fb224588 # v2.1.1
+        uses: peter-evans/repository-dispatch@v2.1.1
         with:
           token: ${{ secrets.WORKFLOW_PAT }}
           repository: "${{github.repository_owner}}/helm-charts"
           event-type: update-chart
-          client-payload: '{"version": "${{ github.event.workflow_run.head_branch }}", "oldVersion": "${{ steps.get_old_release_tag.outputs.release_tag }}", "repository": "${{ github.repository }}"}'
+          client-payload: '{"version": "${{ github.ref_name }}", "oldVersion": "${{ steps.get_last_release_tag.outputs.old_release_tag }}", "repository": "${{ github.repository }}", "crds_asset_id": "${{steps.upload_release_assets.outputs.crds_asset_id}}"}'
```
viccuad, line 130: we should drop `crds_asset_id`, we don't have it in this repo.

Dockerfile
viccuad: I'm ok as it is, resolving.

viccuad (approved):
Approved! Many thanks for all of this! One fix only.

.github/workflows/release.yml
viccuad, line 130:
```suggestion
          client-payload: '{"version": "${{ github.ref_name }}", "oldVersion": "${{ steps.get_last_release_tag.outputs.old_release_tag }}", "repository": "${{ github.repository }}"}'
```

we should drop `crds_asset_id`, we don't have it in this repo.

flavio:
I think we are finally done. This PR makes use of the latest changes introduced inside of the GH actions defined inside of the kubewarden-controller. Moreover, this PR brings back the cross compilation feature.

Straight from b6b8660db364f7abdc32e72a3bc87fc38fca6248 commit:

>     The Dockerfile used for local developement could be used by GitHub
>     actions too, but it turns out to be a non practical solution.
>     
>     GitHub Actions don't have native arm64 runners yet. Because of that the
>     arm64 builds are done using QEMU, which makes everything slow. Building
>     the container inside of a Amazon EC2 ARM instance takes ~8 minutes.
>     Doing that inside of GitHub Actions takes 3 hours and 32 minutes.
>     
>     Because of that, this commits rolls back to performing cross compilation
>     when building the container image inside of a GitHub action. This
>     approach is fast, it takes only 13 minutes to have the ARM container
>     image.
>     
>     However, this PR doesn't goes back to the old cross building solution.
>     It instead uses the `cross-rs` project, which makes everything pretty
>     straightforward. This approach makes possible to perform cross-building
>     also on a local development machine, making easier to reproduce issues.

I've done some experimental builds on my fork:

* push against main: [GH action building the container image](https://github.com/flavio/policy-server/actions/runs/5679729893) takes only 18 minutes to complete
* push of a test release tag: [GH action](https://github.com/flavio/policy-server/actions/runs/5679731944) takes 19 minutes to complete. Note well: the action fails because the fork cannot trigger an update of the helm charts (which is totally fine)
* [Test GH release](https://github.com/flavio/policy-server/releases/tag/v1.7.2):
  * SBOM files are generated for the arm64 and x86_64 images using syft
  * SBOM are attached to the container image inside of the OCI registry and signed

# Notes about cross compilation

The cross compilation is done via [cross-rs](https://github.com/cross-rs/cross), which makes everything incredibly easy. I also tried to do the build on my local machine and it was a painless experience. This is nice, because it allows us to quickly build the policy-server locally, using the same procedure used by GH actions.

cross-rs is also nicely integrated inside of GitHub actions, see [their docs](https://github.com/cross-rs/cross/wiki/FAQ#github-workflows). The integration is done using the `actions-rs/cargo` GH action that we are already using.
There's a "but", the `actions-rs` project is no longer maintained, there are some discussions about what to do next (see https://github.com/actions-rs/toolchain/issues/216). In the future we might have to revisit how we are going to build rust code inside of GH actions. In the worse case scenario, we will have to write our own step doing something like that:

```
run: |
  # install cross via cargo and then...
  cross build --release --target aarch64-unknown-linux-musl
```

.github/workflows/container-image.yml
flavio: Unfortunately they are no longer the same, since now we are back to cross compilation :disappointed:

flavio: Marking as solved, unfortunately there's no solution right now. We have to wait for GitHub to offer native arm64 runners

flavio:
@viccuad can I merge this PR?

viccuad (approved):
LGTM!! Quite a nice change at the end!

PR #518, opened 2023-08-10
feat: policy optimizer

flavio:
This PR introduces a new cli tool called `policy-optimizer`. This binary will download the policies and optimize them. The goal is to implement what is described inside of [this RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0017-policy-server-reuse-wasm-optimized-modules.md).

Currently the code doesn't do any download/optimization. It just leverages the `Lease` primitive declared by Kubernetes to ensure only one process can have write access to the directory where the optimized policies are going to be written.
The actual code similates the download & optimize work with a simple `sleep`.

The change is pretty invasive as you can see, I don't want to merge it in the `main` branch yet.


flavio:
Closing in favor of https://github.com/kubewarden/policy-server/pull/519, which is open against a feature branch of policy-server

PR #536, opened 2023-09-05
chore(deps): be explicit about GH action version

flavio:
Mention all the actions by shasum + their full version.


fabriziosestito (approved):
LGTM

PR #606, opened 2023-11-30
refactor: refactor main function, add lib.rs and refactor config/cli related code

fabriziosestito:
## Description

This PR splits policy server code in lib/bin by adding a `lib.rs` file.
In this way we can call the `policy_server::run` function passing the `Config` struct just introduced by this PR from integration tests in the `tests` folder.

Related to: https://github.com/orgs/kubewarden/projects/6?pane=issue&itemId=44457602


flavio (approved):
Thanks for the cleanup, this looks much better

viccuad (approved):
LGTM!

Cargo.lock
```diff
@@ -3734,6 +3734,7 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "rustls",
+ "rustls-native-certs",
```
viccuad, line 3737: I wonder why this addition, it shouldn't be needed. I suppose cargo will drop it later on.

viccuad: Tried a cargo check of the branch locally, and it gets removed. Not an issue though!

PR #608, opened 2023-12-05
test: add integration tests

fabriziosestito:
## Description

Add policy server integration tests.
The goal is to test that the API are working correctly with the policy evaluator, to then move the runtime related e2e-tests to policy-evaluator.
I've skipped redundant tests (e.g. monitor mode related tests) because we already have unit tests for these scenarios.

## Tradeoff

Unfortunately, due to the complexity of the startup of the worker pool, async-sync world bridge and API server, I wasn't able to use the [warp test utilities
](https://docs.rs/warp/latest/warp/test/index.html).
Right now I am starting the http server and using `reqwest` against it in the integration tests.
If we manage to remove the worker pool, we could try to refactor the code so that it is testable without passing from the http stack.


tests/common/mod.rs
```diff
@@ -0,0 +1,85 @@
+use policy_server::config::{Config, Policy, PolicyMode};
```
fabriziosestito, line 1: we need the old mod.rs here since otherwise `common.rs` will be interpreted as a test by cargo. See: https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests

jvanz (approved):

PR #617, opened 2023-12-19
ci: Fix artifact upload and download

viccuad:
## Description

{upload,download}-artifact v4 don't accept uploading the same named artifact. Suffix their names with the arch, and download all of them at once into the same dir, as it was before.

Relates to https://github.com/kubewarden/policy-server/pull/616.

## Test

Tested on my fork: https://github.com/viccuad/policy-server/actions/runs/7260109787

Release workflow change untested, it's the same as in https://github.com/kubewarden/audit-scanner/pull/162, so refer to that one.


flavio (approved):
Thanks for the fix

PR #627, opened 2024-01-08
Tag 1.10.0-rc1

flavio:
Starting the path that leads to 1.10 release


jvanz (approved):

viccuad:
Should we consume policy-evaluator v0.13.1 in rc1?

viccuad (approved):

Issue #611, opened 2023-12-15
Consider removal of the worker pool 


fabriziosestito:
With the recent memory optimization changes, we realized that removing the worker/worker pool abstraction might be possible by using a semaphore and sharing the EvaluationEnvironment between the API handlers. 
Also, we could consider switching to [axum](https://github.com/tokio-rs/axum)  while refactoring the policy-server.
This allows us to simplify the integration tests using the framework-provided test utilities instead of passing through the HTTP stack.

PR #628, opened 2024-01-10
refactor: remove worker pool

fabriziosestito:
## Description
Fixes: #611 

This PR refactors the policy server by removing the worker thread pool.
Also, it changes the web framework from `warp` to `axum`.

Removing the worker pool is possible because we can now share the `EvaluationEnvironment` between handlers using axum `State` extractor.
A semaphore is used to limit the simultaneous evaluations instead of relying on a pool of workers.
By doing this we can remove the complexity of the worker pool bootstrap and the bridge between sync and async world, also it simplifies the evaluation flow since we do not have to rely on channels and async communication to start an evaluation in the handlers.

This PR also takes care of the following:

- clean-up and major improvements in the layering of the application. The major change is wrapping the policy server code in a PolicyServer type that handles bootstrapping and running the API server.
- adding JSON responses in case of errors (before we were returning plain text and status code)

## Test

Existing tests were moved/updated to comply with the new code.
This PR also takes care of the following:

- adding back timeout protection integration tests that were removed when we removed the venom e2e tests.
- refactoring existing tests using axum testing utilities instead of starting a global HTTP server and sending requests against it.

As expected load testing doesn't show major performance improvements, the old and new implementations show briefly the same results.

Manual metrics/tracing was performed.
TODO: 
- [x] test against a real cluster -> ran e2e tests, passing

## Additional Information

Since the sigstore crate dependes on the blocking `reqwest` feature by using and old version of `tough`, I needed to wrap the fulcio and rekor initialization in a `spawn_blocking` task.
This should go away once we update sigstore-rs, see:
https://github.com/sigstore/sigstore-rs/pull/320

and
https://github.com/kubewarden/policy-server/blob/624ca3b650b2914a4e463a37f53c1dab995e3470/src/lib.rs#L58


src/api/handlers.rs
```diff
@@ -0,0 +1,253 @@
...
+    let response = acquire_semaphore_and_evaluate(
+        state,
+        policy_id,
+        ValidateRequest::AdmissionRequest(admission_review.request),
+        RequestOrigin::Validate,
+    )
+    .await
+    .map_err(handle_evaluation_error)?;
+
+    populate_span_with_policy_evaluation_results(&response);
```
jvanz, line 130: As far as I can see, the tracing for policies in monitor mode will always have result as "allowed". Which is not desired. The metrics should show the original evaluation result. Otherwise, the users will not get the value of the monitor mode which is see the original results before moving the policy to protect mode.

jvanz (changes requested):

src/lib.rs
```diff
@@ -1,407 +1,316 @@
...
-        }
-    };
-    let mem_stats = match process.statm() {
-        Ok(s) => s,
-        Err(e) => {
-            warn!(error =? e, "cannot access process memory stats");
-            return;
-        }
-    };
+        shutdown_tracer_provider();
```
viccuad, line 241: we are calling `shutdown_tracer_provider` here in api_server.run(), but also in main(). Should we only call it in main()?

src/api/handlers.rs
viccuad: I Agree with this. Still this isn't a regression from this PR, so I'm ok opening an issue and tackling it later. I suppose adding a `mode` field and giving value to `allowed` correctly will suffice.

viccuad (approved):

src/api/handlers.rs
flavio: I think we should  create a new issue for that, and tackle that outside of this PR.

I also prefer to keep the current behavior, but extend the trace to have:

- a new field `more` that states whether the policy is operating in `monitor` or `protect` mode
- a new field `raw_result` (or something else) that contains the boolean value of the evaluation result **before** the monitor mode changes that. We could have this field added to all the policies, regardless of their operating mode, or we could have it added only to the traces emitted by policies operating in `monitor` mode

At the end of the day, I want an operator to be able to run a Jaeger query like: `select * from traces where operation_mode = "monitor" and raw_result = false`

src/lib.rs
```diff
@@ -1,407 +1,305 @@
...
+            sigstore::tuf::SigstoreRepository::fetch(None)
+        })
+        .await
+        .unwrap()
+        {
+            Ok(repo) => Some(FulcioAndRekorData::FromTufRepository { repo }),
+            Err(e) => {
+                // We cannot rely on `tracing` yet, because the tracing system has not
+                // been initialized, this has to be done inside of an async block, which
+                // we cannot use yet
```
flavio, line 69: Is that still the case? I think we can switch to tracing now

src/lib.rs
```diff
@@ -1,407 +1,305 @@
...
-                eprintln!(
+        match kube_client {
+            Some(client) => {
+                callback_handler_builder = callback_handler_builder.kube_client(client);
+            }
+            None => {
+                if config.ignore_kubernetes_connection_failure {
+                    // We cannot rely on `tracing` yet, because the tracing system has not
+                    // been initialized yet
+                    eprintln!(
```
flavio, line 103: Same here

flavio:
Fantastic job @fabriziosestito :clap: 

I left some comments, I think we're pretty close to merge this PR

flavio:
Merging, all the concerns/issues have been addressed! :partying_face:

PR #640, opened 2024-01-25
build: Bump version to 1.10.0

viccuad:
## Test

CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3612,7 +3612,7 @@ dependencies = [
 
 [[package]]
 name = "policy-server"
-version = "1.10.0-rc2"
+version = "1.10.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


flavio:
Ouch, I've nothing against that... but this would cause all the changes that got into `main` to be part of the final 1.10.0 release

flavio:
This is going to have the final 1.10.0 release to consume latest versions of k8s-openapi and kube.

I think this is fine, everything tested green inside of the unit, integration and e2e tests

However, just to be safe... can you please rebase this PR against main? I literally merged the dep update 2 seconds ago.

If everything is green after the rebase, go ahead and merge it

viccuad:
Rebased onto main, let's see!

Issue #661, opened 2024-02-06
policy-server: Enable codecov


viccuad:
Port the work done in https://github.com/kubewarden/policy-evaluator/pull/434 for policy-server.

viccuad:
tarpaulin is failing to build the policy-server bin, both with `stable` and `nightly`, and with `ptrace` and `llvm` engines. I keep working on this.

PR #667, opened 2024-02-15
ci: Add codecov support

viccuad:
## Description

Fix https://github.com/kubewarden/policy-server/issues/661

Add new make targets `coverage, `coverage-unit-tests`, `coverage-integration-tests` that are used in the CI. They run the tests via `cargo-tarpaulin` with llvm, and save the reports to the respective new folders `coverage/{unit-tests,integration-tests}/*`.

The make targets are both configured with `cargo +nightly`, as `cargo-tarpaulin` for the integration-tests need it, and in this way we can skip recompiling again. 

Add new CI job that runs the tests via the new makefile targets and uploads the result to codecov.

## Test

CI, PR should get a "welcome to codecov" message, and audit-scanner appear under https://app.codecov.io/github/kubewarden/.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,3 +84,39 @@ jobs:
         with:
           command: clippy
           args: -- -D warnings
+
+  coverage:
+    name: coverage
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


flavio:
nice job!

Issue #682, opened 2024-02-29
Feature Request: do not exit with an error if one of the policies cannot be instantiated


flavio:
### Is your feature request related to a problem?

This is a change require to address a problem of the Policy Server.

Currently, if one of the defined policies cannot be loaded (maybe it cannot be downloaded from the remote registry, maybe the user provided settings that are not valid,...) the Policy Server process exits with an error.

Inside of Kubernetes, the Pod running the Policy Server will be restarted a bunch of time and then it will be left in crash loop state. The only way to recover from this situation is to have someone look into the error message of the Policy Server Pod and fix the issue.

This behavior is dangerous. When rolling up a new policy (or making any change to the existing ones), the new Policy Server Pods could be end up in this broken state. The old ones, still running with the old working configuration, are going to disappear if something happens to the node where they are scheduled.

Because of that, it's possible to end up with a broken cluster: all the incoming admission requests are rejected because there are no working instances of Policy Server.

### Solution you'd like

Instead of exiting with an error, Policy Server should boot regularly. All the requests sent to the endpoint owned by the broken policy must return a rejection message and provide a meaningful explanation of what went wrong with the policy.

It would also be great to have the error message added to the events of the `ClusterAdmissionPolicy`/`AdmissionPolicy` associated.

flavio:
> It would also be great to have the error message added to the events of the ClusterAdmissionPolicy/AdmissionPolicy associated.

This is going to be covered later, as part of another feature.

**TODO:** create an issue that keeps track of this separate task

PR #712, opened 2024-03-19
feat: do not exit on policy error

fabriziosestito:
## Description

Fix: #682

The policy server was exiting with an error If a policy could not be downloaded, verified, or had invalid settings, causing the Pod to restart. The policy server pod entered a crash back-off loop in the worst-case scenario. 
This PR changes the way the policy server manages policy initialization errors by not exiting and storing the error in the evaluation environment so it can be bubbled up to the API in the form of a rejected `AdmissionResponse` with the error as message and the status code set to `500`.

### DIFF
--- a/src/evaluation/precompiled_policy.rs
+++ b/src/evaluation/precompiled_policy.rs
@@ -66,8 +66,9 @@ impl PrecompiledPolicy {
 
 /// A dictionary with:
 /// * Key: the URL of the WebAssembly module
-/// * value: the PrecompiledPolicy
-pub(crate) type PrecompiledPolicies = HashMap<String, PrecompiledPolicy>;
+/// * Value: a Result containing the precompiled policy or an error.
+/// Errors are stored and will be reported to the user in the API response.
+pub(crate) type PrecompiledPolicies = HashMap<String, Result<PrecompiledPolicy>>;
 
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -165,7 +164,7 @@ impl PolicyServer {
             info!("policy timeout protection is disabled");
         }
 
-        verify_policy_settings(&config.policies, &evaluation_environment).await?;
+        // verify_policy_settings(&config.policies, &evaluation_environment).await?;
```
jvanz, line 167: Remove this commented line. ;)

jvanz (approved):

viccuad (approved):
LGTM modulo José's comment!

src/evaluation/precompiled_policy.rs
```diff
@@ -67,7 +67,7 @@ impl PrecompiledPolicy {
 /// A dictionary with:
 /// * Key: the URL of the WebAssembly module
 /// * value: the PrecompiledPolicy
```
flavio, line 69: Please update the inline doc, now this is a `Result<X>`, explain why we could be storing an `Err`

flavio (changes requested):
LGTM, we can merge it as soon as the two changes requested are done (mine + the one from @jvanz)

flavio (approved):

PR #713, opened 2024-03-20
feat: add policy initialization error metric

fabriziosestito:
## Description

Adds policy initialization error metric


flavio (approved):
LGTM

Issue #730, opened 2024-04-12
feature: introduce feature flag that influences how policy server reacts to broken policies


flavio:
Policy Server quits when one of its policies cannot be downloaded or when the policy configuration is not valid. This has always been its behavior.

We recently introduced a change that causes policy server to not exit when one of these circumstances take place. Instead, the policy server keeps running but the requests sent to one of the broken policies are rejected with a special error message.

This can be confusing and lead to some unexpected scenarios.

Because of that we want the following changes to happen.

## Acceptance criteria

- [ ] A new feature flag + env variable (boolean one) is added to policy server. This determines if the policy server should exit on error or keep running
- [ ] By default the policy server should retain the old behavior
- [ ] We have to document this new feature flag, explaining why we did it and its limitations
- [ ] Advertise this feature as alpha, look for feedback from users

flavio:
@fabriziosestito we've assigned that to you since you've already looked into this topic

PR #733, opened 2024-04-15
fix: add hidden flag/env to toggle continuing on initialization errors

fabriziosestito:
## Description

Adds a flag to toggle the "continue on errors" behavior, so we can keep working on the policy server error handling flow without breaking the current user experience.

Fixes: #730

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -137,6 +137,14 @@ impl PolicyServer {
         let engine = wasmtime::Engine::new(&wasmtime_config)?;
         let precompiled_policies = precompile_policies(&engine, &fetched_policies);
 
+        if !config.continue_on_errors {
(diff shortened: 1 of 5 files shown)


jvanz (approved):

src/lib.rs
```diff
@@ -137,13 +137,22 @@ impl PolicyServer {
         let engine = wasmtime::Engine::new(&wasmtime_config)?;
         let precompiled_policies = precompile_policies(&engine, &fetched_policies);
 
+        if !config.continue_on_errors {
+            for result in precompiled_policies.values() {
```
flavio, line 141: I think it would be nice to collect all the error messages and then return them as a single one. In this way the user would know about all the errors, instead of just the 1st one

PR #737, opened 2024-04-16
fix: sigstore cache dir error

### LINKED ISSUE #736: Error: No such file or directory (os error 2) when starting PolicyServer

### Current Behavior

After merging sigstore related fixes, we have Error: No such file or directory (os error 2).
This is caused by the sigstore cache directory not being found, see https://github.com/kubewarden/policy-server/blob/30c2cdb7efec01201342aa41579296b8bccbad29/src/lib.rs#L66

### Expected Behavior

_No response_

### Steps To Reproduce

Start the latest policy server 

### Environment

_No response_

### Anything else?

_No response_

fabriziosestito:
## Description

Fixes #736 

## Additional Information

`PolicyDownloader` was refactored so that we initialize the `ManualTrustRoot` just once and pass it to the constructor.

### DIFF
--- a/src/policy_downloader.rs
+++ b/src/policy_downloader.rs
@@ -8,10 +8,9 @@ use policy_evaluator::{
     },
     policy_metadata::Metadata,
 };
-use sigstore::trust::{ManualTrustRoot, TrustRoot};
+use sigstore::trust::ManualTrustRoot;
 use std::{
     collections::{HashMap, HashSet},
-    fs,
     path::{Path, PathBuf},
     sync::Arc,
 };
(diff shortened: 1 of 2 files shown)


src/policy_downloader.rs
```diff
@@ -340,8 +314,31 @@ mod tests {
...
+        let manual_root = ManualTrustRoot {
+            fulcio_certs: Some(fulcio_certs),
+            rekor_keys: Some(
+                repo.rekor_keys()
+                    .expect("Cannot fetch Rekor keys from TUF repository")
+                    .iter()
+                    .map(|k| k.to_vec())
+                    .collect(),
+            ),
+        };
```
fabriziosestito, line 337: we do the same in the PolicyServer constructor, should we create a utility function in another module?

viccuad: I'm ok with not doing it, the upstream types and functions are bound to scream at us if there's a change, and there is no logic.

viccuad (approved):
LGTM. Maybe a test for `new_from_config()` would be good, but mocking it seems quite complicated for essentially testing only that the dir is created and populated if it didn't exist.

src/policy_downloader.rs
flavio: I would leave it as it currently is, I agree with @viccuad

flavio (approved):
@fabriziosestito thanks a lot for having fixed it! :heart:

fabriziosestito:
> LGTM. Maybe a test for `new_from_config()` would be good, but mocking it seems quite complicated for essentially testing only that the dir is created and populated if it didn't exist.

@viccuad we are testing `new_from_config` indirectly in the integration tests, maybe we could extend those tests with a signed policy. Since we test `PolicyDownloader` in isolation, they will be redundant, so I am not sure it is worth it.

PR #768, opened 2024-05-21
feat: support heap profiling

flavio:
Support the generation of pprof profiles of heap usage by using jemalloc facilities.

This requires jemalloc to be configured with some special options, hence all the builds will have this settings enabled. This doesn't seem to cause performance regressions from what I've seen while doing load tests locally.

I've also added some instructions explaining how to memory and CPU profiling.


jvanz (approved):
The changes looks good to me. But my personal preference is to enable this feature only when required using build time configs.

flavio:
I've done the following changes:

- jemalloc's pprof support is enabled by default, but is NOT active
- when jemalloc's pprof is active there's a higher CPU usage
- jemalloc's pprof is automatically activated once the 1st heap profile generation is requested
- once activated, the pprof cannot be deactivate. I could change it, but that would require adding a new API endpoint. It's not then end of the world, but I don't think we need it

fabriziosestito (approved):

PR #770, opened 2024-05-22
fix: configure jemalloc in lib.rs

fabriziosestito:
## Description

Integration tests were failing since we were activating the profiling, but there jemalloc was not configured in `lib.rs` (integration tests instantiate a policy server using the library part of the crate).

Also, we don't need to enable pprof when testing, since it could slow down the CI.


flavio (approved):
good catch!

PR #773, opened 2024-05-22
test: fix permissions of cargo audit action

flavio:
This should prevent the action from failing


flavio:
Merging, this fixes the issue. This has been tested on the kwctl repository

PR #792, opened 2024-06-13
sigstore fixes

flavio:
Ensure sigstore Trust Root is always fetched. We must always fetch the Sigstore Trust Root, otherwise policies making use of Sigstore won't be able to do keyless verifications.

Prior to this commit, the Trust Root was fetched only when the policy integrity verification was turned on.


viccuad (approved):
LGTM.

fabriziosestito (approved):

PR #819, opened 2024-07-09
deps: Bump opentelemetry dependencies

viccuad:
## Test

- CI.
- Performing docker-compose local deployment and testing: Tracing is broken, as jaeger fails to connect with otel-collector. This is because we aren't pinning jaeger.
- end-to-end tests: broken for latest jaeger-operator-2.54.0. Working for jaeger-operator-2.49.0, the one we list as supported in https://docs.kubewarden.io/reference/dependency-matrix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -936,15 +936,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-channel"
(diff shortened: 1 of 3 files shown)


flavio (approved):
Thanks for the fixes!

PR #826, opened 2024-07-16
feat: automatic TLS certificate reload

flavio:
When running on Linux, monitor changes done to the TLS certificate and key used by the https server.

When both change, react by updating the TLS configuration.

This allows to rotate the TLS certificate used by a Policy Server without having to restart the process.

Notes:

- This is not covered by the unit tests right now. I don't know yet how hard it would be to test it. I'll leave a comment here once I know more about the topic
- This change is part of the upcoming work to handle certificate rotation inside of the whole KW stack
- Ideally, we can merge this code right now and make it part of the 1.15 release since it won't cause any harm

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -493,7 +493,7 @@ checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 [[package]]
 name = "burrego"
 version = "0.3.4"
-source = "git+https://github.com/kubewarden/policy-evaluator?tag=v0.18.1#faea40d47f9fc91728663715aac1fb3131b53f30"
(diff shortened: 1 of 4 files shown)


fabriziosestito (approved):
Looks great! As far as testing goes, we could write an integration test. However, I don't see this as critical and will be fine leaving this part untested.

jvanz (approved):

flavio:
I've updated the code to have an integration test. The integration test will fail until I do the following changes:
- merge https://github.com/kubewarden/policy-fetcher/pull/201 and tag a new release of policy-fetcher
- tag a new release of policy-evaluator that consumes this version of policy-fetcher

jvanz (approved):

viccuad (approved):
LGTM! :rocket:

PR #827, opened 2024-07-17
fix policies yml parsing

flavio:
Address the failures reported by our CI system.

They were caused by two changes done to the `policies.yml` file parsing:

- The usage of `deny_unknown_fields` with untagged `Enum` is affected by a bug inside of the `serde` library. This is a known issue
- The `policies.yml.example` file had invalid contents


viccuad (approved):
LGTM. Would prefer a unit test that catches this though.

jvanz (approved):

fabriziosestito (approved):

PR #843, opened 2024-07-24
fix: policies.yml parsing

fabriziosestito:
## Description

Due to the recent addition of policy groups, the parsing of policies.yml was broken.
This happened because of [a pretty confusing serde behavior](https://github.com/serde-rs/json/issues/501), as an attribute on an enum is renaming all the variants of the enum, not all the fields of those variants.

TODO:
- [x] update tests

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -258,7 +258,7 @@ fn remote_server_options(matches: &clap::ArgMatches) -> Result<Option<Sources>> 
     Ok(sources)
 }
 
-#[derive(Deserialize, Debug, Clone, Default)]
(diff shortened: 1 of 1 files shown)


jvanz:
I think this PR fixes the `v1.15.0-rc1` issue we are discussing in other channels. I was testing the bug and when I applied the same change, the problem goes away. Serde was ignoring the policy mode field due the funky behavior that you linked here. To double check that, I've renamed the field to "policy_mode" in my testing policies.yml and that works as well. :slightly_smiling_face:

viccuad (approved):
LGTM!

Issue #855, opened 2024-08-02
Add opentelemetry integration tests


viccuad:
### Acceptance criteria

Opentelemetry integration with metrics and tracing is tested as part of integration tests, by hitting the metrics and tracing endpoints.

PR #859, opened 2024-08-06
test: add otelc integration test

fabriziosestito:
## Description
Adds otel collector integration test.

The otelc container is started using  `testcontainers`, and configured with file exporters.
We can then read and parse the file contents and assert that everything is fine.

Fixes: #855

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,8 @@ rstest = "0.21"
 tempfile = "3.10.1"
 tower = { version = "0.4", features = ["util"] }
 http-body-util = "0.1.1"
+testcontainers = { version = "0.21.1", features = ["watchdog"] }
+backon = "0.4.4"
 
 [target.'cfg(target_os = "linux")'.dev-dependencies]
 rcgen = { version = "0.13", features = ["crypto"] }
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -1,6 +1,14 @@
 mod common;
 
-use std::collections::{BTreeSet, HashMap};
+use std::io::BufReader;
+use std::{
+    collections::{BTreeSet, HashMap},
+    fs::{set_permissions, File, Permissions},
+    io::BufRead,
+    os::unix::fs::PermissionsExt,
+    path::PathBuf,
+    time::Duration,
+};
 
 use common::app;
 
@@ -8,6 +16,7 @@ use axum::{
     body::Body,
     http::{self, header, Request},
 };
+use backon::{ExponentialBuilder, Retryable};
 use http_body_util::BodyExt;
 use policy_evaluator::{
     admission_response::AdmissionResponseStatus,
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -65,6 +65,8 @@ rstest = "0.21"
 tempfile = "3.10.1"
 tower = { version = "0.4", features = ["util"] }
 http-body-util = "0.1.1"
+testcontainers = { version = "0.21.1", features = ["watchdog"] }
+backon = "0.4.4"
```
flavio, line 69: Nice addition! I think I can refactor the unit test of the certificate reload to make use of it

tests/integration_test.rs
```diff
@@ -16,12 +24,21 @@ use policy_evaluator::{
...
+use tempfile::NamedTempFile;
+use testcontainers::{
+    core::{Mount, WaitFor},
+    runners::AsyncRunner,
+    GenericImage, ImageExt,
+};
 use tower::ServiceExt;
 
 use crate::common::default_test_config;
+use backon::{ExponentialBuilder, Retryable};
```
flavio, line 41: nitpick: should be moved upper, like around line 38

tests/integration_test.rs
```diff
@@ -625,7 +642,8 @@ async fn test_detect_certificate_rotation() {
     let key_file = certs_dir.path().join("policy-server-key.pem");
 
     let hostname1 = "cert1.example.com";
-    let tls_data1 = create_cert(&hostname1);
+    let var_name = &hostname1;
+    let tls_data1 = create_cert(var_name);
```
flavio, line 646: I'm a bit puzzled about this change, nothing else was touched inside of the test. Do we need it?

tests/integration_test.rs
```diff
@@ -679,3 +697,105 @@ async fn test_detect_certificate_rotation() {
...
+    assert_eq!(spans["spans"][0]["name"], "policy_log");
+    assert_eq!(spans["spans"][1]["name"], "validate_settings");
+
+    otelc.stop().await.unwrap();
+}
+
+async fn parse_exporter_output(
+    exporter_output_file: &File,
+) -> serde_json::Result<serde_json::Value> {
+    // let exporter_output_file = File::open(exporter_output_file_path).unwrap();
```
flavio, line 793: Can be removed

flavio (changes requested):
Looks really neat! I left some minor comments, once addressed this is ready to be merged

tests/integration_test.rs
fabriziosestito: the linter was complaining, i just applied the suggestion

fabriziosestito: you are right, it seems i've applied the wrong code action, let me revert it

viccuad (approved):
LGTM!

Cargo.toml
jvanz: Indeed! Very cool

jvanz (approved):

flavio (approved):

PR #863, opened 2024-08-09
fix: properly configure rustls stack

flavio:
Starting from the 0.22 release, rustls requires each application to configure its default crypto provider. Without that, the application will face a runtime panic.

This is something the "cert-reload" integration test allowed us to expose.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,6 +12,12 @@ use policy_server::PolicyServer;
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    // Starting from rustls 0.22, each application must set its default crypto provider.
+    let crypto_provider = rustls::crypto::aws_lc_rs::default_provider();
+    crypto_provider
+        .install_default()
+        .expect("Failed to install crypto provider");
+
     let matches = cli::build_cli().get_matches();
     let config = policy_server::config::Config::from_args(&matches)?;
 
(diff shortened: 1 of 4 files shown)


flavio:
@viccuad this is the fix for the issue you've been looking into. The axum-server update caused a bump of the rustls crate, which lead to this issue.

src/main.rs
```diff
@@ -12,6 +12,12 @@ use policy_server::PolicyServer;
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    // Starting from rustls 0.22, each application must set its default crypto provider.
+    let crypto_provider = rustls::crypto::aws_lc_rs::default_provider();
```
viccuad, line 16:
```suggestion
    let crypto_provider = rustls::ring::aws_lc_rs::default_provider();
```

`aws_lc_rs` needs additional dependencies on Windows (not relevant here), yet AFAIK we were going with ring everywhere.

tests/integration_test.rs
```diff
@@ -637,6 +637,15 @@ mod certificate_reload_helpers {
 async fn test_detect_certificate_rotation() {
     use certificate_reload_helpers::*;
 
+    // Starting from rustls 0.22, each application must set its default crypto provider.
+    // This setup is done inside of the `main` function of the policy server,
+    // which is not called in this test.
+    // Hence we have to setup the crypto provider here.
+    let crypto_provider = rustls::crypto::aws_lc_rs::default_provider();
```
viccuad, line 644:
```suggestion
    let crypto_provider = rustls::ring::aws_lc_rs::default_provider();
```

viccuad (approved):

flavio:
I've switched to `ring` as crypto provider. I'll merge the PR once everything is green

Issue #846, opened 2024-07-25
Building with musl target fails locally


fabriziosestito:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Building with musl target fails locally.
Both the `Dockerfile` and `cargo build --release --target=x86_64-unknown-linux-musl` fail with the same error:
```
  = note: /sbin/ld: /usr/lib/libm-2.40.a(s_truncf.o): warning: relocation against `_dl_x86_cpu_features' in read-only section `.text'
          /sbin/ld: /usr/lib/libm-2.40.a(w_exp.o): in function `__exp':
          (.text+0x62): undefined reference to `errno'
          /sbin/ld: /usr/lib/libm-2.40.a(w_log.o): in function `__log':
          (.text+0x1f): undefined reference to `errno'
          /sbin/ld: /usr/lib/libm-2.40.a(w_pow.o): in function `__pow':
          (.text+0x7a): undefined reference to `errno'
          /sbin/ld: (.text+0xb8): undefined reference to `errno'
          /sbin/ld: /usr/lib/libm-2.40.a(e_exp.o): in function `__ieee754_exp_ifunc':
          (.text+0x236): undefined reference to `_dl_x86_cpu_features'
          /sbin/ld: (.text+0x241): undefined reference to `_dl_x86_cpu_features'
          /sbin/ld: (.text+0x24a): undefined reference to `_dl_x86_cpu_features'
          /sbin/ld: /usr/lib/libm-2.40.a(e_log.o): in function `__ieee754_log_ifunc':
          (.text+0x2c6): undefined reference to `_dl_x86_cpu_features'
          /sbin/ld: (.text+0x2d1): undefined reference to `_dl_x86_cpu_features'
          /sbin/ld: /usr/lib/libm-2.40.a(e_log.o):(.text+0x2da): more undefined references to `_dl_x86_cpu_features' follow
          /sbin/ld: /usr/lib/libm-2.40.a(math_err.o): in function `with_errno':
          (.text+0x3): undefined reference to `errno'
          /sbin/ld: read-only segment has dynamic relocations
          collect2: error: ld returned 1 exit status
```

However, the CI build job is working fine.

This seems to be related to: https://github.com/bytecodealliance/wasmtime/issues/8898
[...]

PR #874, opened 2024-08-13
fix: allow Docker image to be built locally

flavio:
Fix building the docker image locally.

Addresses https://github.com/kubewarden/policy-server/issues/846


viccuad (approved):
Tested locally, works fine. Approving.

Maybe we should keep the issue around still, as direct builds will not work.

PR #908, opened 2024-09-18
group policy: improve evaluation response

flavio:
When evaluating group policy, tell the user which sub-policies have not been evaluated because their outcome did not matter for the final result.


src/evaluation/evaluation_environment.rs
```diff
@@ -800,16 +844,16 @@ mod tests {
...
-                        url: "file:///tmp/policy_1.wasm".to_string(),
+                        url: "file:///tmp/happy_policy_1.wasm".to_string(),
                         settings: None,
                         context_aware_resources: BTreeSet::new(),
                     },
                 )]
                 .into_iter()
                 .collect(),
-                expression: "not_a_known_policy() || policy_1()".to_string(),
+                expression: "not_a_known_policy() || happy_policy_1()".to_string(),
```
fabriziosestito, line 856: nit: `unknown_policy`

fabriziosestito (approved):

src/evaluation/evaluation_environment.rs
flavio: you're right, fixed

PR #909, opened 2024-09-19
fix: revisit AdmissionResponse of policy groups

flavio:
The API Server puts some limitations on the warnings:

- they cannot exceed 256 characters
- the size of all the warnings cannot exceed 4096 characters
- they are returned as HTTP headers, hence not all characters are allowed

Because of these reasons, starting from now, we use the warning struct only to tell the user whether a member policy was evaluated or not. When it was evaluated we just tell the outcome (allow/reject).

The details of each policy evaluation are returned as part of the `AdmissionResponse.status.details.causes`.

Notes: the warning messages are always shown, regardless of the request being accepted or rejected.
The `AdmissionResponse.status.details.causes` are shown only when the request is rejected. `kubectl` shows them only when run with verbose major or equal to 4 (`-v4`).

Policy Server always shows the `causes` inside of its logs when running in `debug` mode.

The `causes` are always sent as part of the traces emitted by Policy Server.

### DIFF
--- a/src/evaluation/evaluation_environment.rs
+++ b/src/evaluation/evaluation_environment.rs
@@ -1,5 +1,5 @@
 use policy_evaluator::{
-    admission_response::{AdmissionResponse, AdmissionResponseStatus},
+    admission_response::{self, AdmissionResponse, AdmissionResponseStatus},
     callback_requests::CallbackRequest,
     evaluation_context::EvaluationContext,
     kubewarden_policy_sdk::settings::SettingsValidationResponse,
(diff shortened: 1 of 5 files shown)


src/evaluation/evaluation_environment.rs
```diff
@@ -659,28 +659,70 @@ impl EvaluationEnvironment {
...
+        let mut warnings = vec![];
+
+        // The details of each policy evaluation are returned as part of the
+        // AdmissionResponse.status.details.causes
+        let mut status_causes = vec![];
+
+        let evaluation_results = policies_evaluation_results.lock().unwrap();
+
+        for policy_id in &policy_ids {
+            match evaluation_results.get(policy_id) {
```
fabriziosestito, line 679: nit: we could use `if let` to reduce indentation

fabriziosestito (approved):

fabriziosestito:
Overall LGTM, some integration tests are failing due to the updated `AdmissioResponse`

src/evaluation/evaluation_environment.rs
flavio: done :+1:

flavio:
everything should be green now, I'll merge the PR once everything passes

PR #924, opened 2024-09-25
fix(integration tests): cannot run single integration test

fabriziosestito:
## Description
While investigating an flaky integration tests, I needed to run the test in isolation.
The test was failing with: 
```
no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is caused by the default crypto provider not being installed.
The integration tests are working as expected because the TLS hot-reload test installs a default crypto provider, so the other tests work. 

This PR adds a function using `std::sync::Once` as a global `setup` for integration tests.
The `setup` function sets the crypto default provider. In the future, it could be used to share a common setup that needs to be executed once for all the test suite.

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -9,6 +9,23 @@ use std::{
 };
 use tempfile::tempdir;
 
+use std::sync::Once;
+static START: Once = Once::new();
+
+/// Common setup for tests. This function should be called at the beginning of each test.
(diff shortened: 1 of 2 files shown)


flavio (approved):
I like it, thanks for the fix

Issue #971, opened 2024-11-11
Policy Server SBOM is empty


flavio:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I've looked into the SBOM generated by latest release (1.18.0) and there's no relevant data in there, it's basically empty.

### Expected Behavior

The SBOM should include all the information about the dependencies and tools used to build the artifact.

### Steps To Reproduce

There are two ways to get the SBOM. The first one is from the [1.18.0 release page](https://github.com/kubewarden/policy-server/releases/tag/v1.18.0), download the `attestation-<platform>.tar.gz` file, unpack it and look at the contents of the SBOM file.

The second one is by using the [`slsactl`](https://github.com/rancherlabs/slsactl/) CLI tool created by the Rancher Security team:

```console
slsactl download sbom ghcr.io/kubewarden/policy-server:v1.18.0
```

Compare the output produced by the previous command with the following ones:

```console
slsactl download sbom ghcr.io/kubewarden/kubewarden-controller:v1.18.0
slsactl download sbom ghcr.io/kubewarden/audit-scanner:v1.18.0
```

### Environment

```markdown
Kubewarden 1.18.0 release
```

### Anything else?

There's definitely something wrong with the SBOM generation for Policy Server, maybe this is related with Rust or with the usage of cross-rs... :thinking: 

The SBOMs of `kubewarden-controller` and `audit-scanner` are good

pjbgf:
The default [Docker Buildx backend](https://github.com/docker/buildkit-syft-scanner) for sbom generation is `syft`, which currently fails to identify any package on that image and don't seem to be able to do composition analysis for the Rust executable:
```
syft ghcr.io/kubewarden/policy-server:v1.18.0 --scope=all-layers
 ✔ Loaded image                                                       ghcr.io/kubewarden/policy-server:v1.18.0
 ✔ Parsed image                        sha256:e348d2e67de2de6509f4f2787a96234caed896b37c273c6132bc3877bf41db1b
 ✔ Cataloged contents                         edf561b8f457e89ed2adeae9e4447184d4344b2080a10088f38c9e53049e2ea4
   ├── ✔ Packages                        [0 packages]
   └── ✔ Executables                     [1 executables]
No packages discovered
```

Which results in an empty SBOM.

PR #972, opened 2024-11-12
fix: make policy-server binary auditable

flavio:
When creating the policy-server image, build using `cargo auditable`.
This will enrich the `policy-server` binary with its bill of material.

Fixes https://github.com/kubewarden/policy-server/issues/971

## Testing

I tested the fix by doing a local build of the policy-server image and then
running `syft` against the image.

The output I got was complete.


Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN mkdir /usr/src/policy-server
 WORKDIR /usr/src/policy-server
 COPY ./ ./
 
-RUN cargo install --target aarch64-unknown-linux-musl --path .
+RUN cargo install cargo-auditable
+RUN alias cargo="cargo auditable" && cargo install --target aarch64-unknown-linux-musl --path .
```
jvanz, line 17: Why not this?:

```suggestion
RUN cargo install cargo-auditable
RUN cargo auditable install --target aarch64-unknown-linux-musl --path .
```

viccuad (approved):
LGTM, I'm ok both with alias or without.

Dockerfile
flavio: I didn't expect that to work, based on some GH issue I found, but it actually works!

I've changed the code as you suggested

flavio:
I changed the code to not use the `alias`, it turned out this is not needed if you call `cargo auditable install`

PR #978, opened 2024-11-14
chore(deps): updates OpenTelemetry dependencies

jvanz:
## Description

Updates policy-server to use the latest version of OpenTelemetry libraries.


jvanz:
@kubewarden/kubewarden-developers I've ran the e2e tests using a policy server image with this changes. They're all green. I think you can review it as well.

flavio (approved):

Issue #993, opened 2024-12-04
Support send data to remote Otel collector


jvanz:
This issue is keep track of a subset of the changes necessary to allow Kubewarden send data to remote OpenTelemetry (Otel) collectors. As described in https://github.com/kubewarden/kubewarden-controller/issues/931.

Currently, the policy server is able to send data to a Otel collector running as sidecar. It's necessary to change exporter in the code to send data to a remote Otel collector.

### Acceptance criteria
- [x] Change the metrics/tracing exporter to consider the environment variables to configure the Otel exporter.
- [ ] Add configuration to store the location of the certificate to be used in the secure communication with Otel collector
  - Consider contributing upstream in the OpenTelemetry rust library to load the certificate path also from environment variable. As it's done in the Go lang libraries

fabriziosestito:
The configuration will initially be implemented in the policy server using CLI flags or environment variables. Once completed, contributions will be made upstream, as the upstream merge process might take some time.

PR #994, opened 2024-12-05
feat: add otlp endpoint config and cli flag

fabriziosestito:
## Description

This PR adds a new `--otlp-endpoint` flag and the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable, allowing users to specify a custom gRPC OTLP endpoint for exporting both metrics and traces.

Part of #993 

## Test

Updates existing integration tests.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -169,7 +169,12 @@ pub(crate) fn build_cli() -> Command {
                 .long("continue-on-errors")
                 .env("KUBEWARDEN_CONTINUE_ON_ERRORS")
                 .action(ArgAction::SetTrue)
-                .hide(true)
+                .hide(true),
+            Arg::new("otlp-endpoint")
+                .long("otlp-endpoint")
+                .env("OTEL_EXPORTER_OTLP_ENDPOINT")
+                .default_value("http://localhost:4317")
+                .help("The OTLP gRPC endpoint for exporting traces and metrics.")
     ];
     args.sort_by(|a, b| a.get_id().cmp(b.get_id()));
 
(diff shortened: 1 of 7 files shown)


jvanz (changes requested):
Unless we really want a CLI flag for this, I think we do not need it. The rust library does [check the environment variable for us](https://github.com/open-telemetry/opentelemetry-rust/blob/957659fac9217534e64d36f80bddc8360345fde5/opentelemetry-otlp/src/exporter/tonic/mod.rs#L191). 

I would add a CLI flag to define the certificate now instead. ;)

fabriziosestito:
@jvanz I think we should keep both the CLI flag and the config entry. It’s better to have consistent configurations, and exposing the certificate path only through the CLI could be confusing. Having the config struct map the otlp CLI values also makes integration tests easier.

src/cli.rs
```diff
@@ -169,7 +169,12 @@ pub(crate) fn build_cli() -> Command {
                 .long("continue-on-errors")
                 .env("KUBEWARDEN_CONTINUE_ON_ERRORS")
                 .action(ArgAction::SetTrue)
-                .hide(true)
+                .hide(true),
+            Arg::new("otlp-endpoint")
+                .long("otlp-endpoint")
+                .env("OTEL_EXPORTER_OTLP_ENDPOINT")
```
fabriziosestito, line 175: we could change this to `KUBEWARDEN_OTLP_ENDPOINT`if we decide to keep this cli flag

jvanz:
> @jvanz I think we should keep both the CLI flag and the config entry. It’s better to have consistent configurations, and exposing the certificate path only through the CLI could be confusing. Having the config struct map the otlp CLI values also makes integration tests easier.

Fair enough. I'm fine with that. I was mentioning that just to ensure that you're aware of this possibility. ;)

jvanz (approved):

flavio (approved):
LGTM

PR #998, opened 2024-12-09
feat: add otlp tls configuration

### LINKED ISSUE #993: Support send data to remote Otel collector

fabriziosestito:
## Description
This PR adds the following flags and env variables:

- `otlp-certificate`/`OTEL_EXPORTER_OTLP_CERTIFICATE`:  The trusted certificate to use when verifying a server's TLS credentials.  
- `otlp-client-certificate`/`OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE`: Client certificate/chain trust for clients private key to use in mTLS communication in PEM format.
- `otlp-client-key`/ `OTEL_EXPORTER_OTLP_CLIENT_KEY`: Client certificate/chain trust for clients private key to use in mTLS communication in PEM format.

## Test
Updates otel integration test

### Tradeoff

Rust opentelemetry only supports tls by using tonic (grpc).

Closes #993


src/cli.rs
```diff
@@ -173,8 +173,23 @@ pub(crate) fn build_cli() -> Command {
             Arg::new("otlp-endpoint")
                 .long("otlp-endpoint")
                 .env("OTEL_EXPORTER_OTLP_ENDPOINT")
-                .default_value("http://localhost:4317")
-                .help("The OTLP gRPC endpoint for exporting traces and metrics.")
+                .help("The OTLP gRPC endpoint for exporting traces and metrics."),
```
fabriziosestito, line 176: memo to self/@fabriziosestito: this is a wip fix flags

tests/data/otel-collector-config.yaml
```diff
@@ -5,6 +5,10 @@ receivers:
   otlp:
     protocols:
       grpc:
+        tls:
+          ca_file: "certs/ca.pem"
+          cert_file: "certs/server-cert.pem"
+          key_file: "certs/server-key.pem"
```
jvanz, line 11: I'm not sure if this is testing mTLS:

```
Receivers leverage server configuration.

Beyond TLS configuration, the following setting can optionally be configured (required for mTLS):

    - client_ca_file: Path to the TLS cert to use by the server to verify a client certificate. (optional) This sets the ClientCAs and ClientAuth to RequireAndVerifyClientCert in the TLSConfig. Please refer to https://godoc.org/crypto/tls#Config for more information.
    - client_ca_file_reload (default = false): Reload the ClientCAs file when it is modified.
```

https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md#server-configuration

Is this in purpose?

fabriziosestito: you are right, it's missing the client ca file

flavio (commented):
Everything looks good, modulo what Jose pointed out about the test.

Moreover, as a nitpick, I would not add the certificate CA, keys and certificates to the test directory, but I would generate them before running the tests. This can be easily done with the [rcgen](https://docs.rs/rcgen/latest/rcgen/) crate from the rustls project.

fabriziosestito:
@flavio @jvanz I've added a new commit https://github.com/kubewarden/policy-server/pull/998/commits/8a011af2798f221df790923f9e3512254ec3cfc3 which removes the flags and uses the env variables for everything.
However, I needed to implement the certificates env vars logic which is not provided by `opentelemetry-rust`, see https://github.com/open-telemetry/opentelemetry-rust/issues/984

tests/integration_test.rs
```diff
@@ -746,18 +748,50 @@ async fn test_detect_certificate_rotation() {
...
+    let client_cert_file = NamedTempFile::new().unwrap();
+    let client_key_file = NamedTempFile::new().unwrap();
+
+    let files_and_contents = [
+        (server_ca_file.path(), &server_ca),
+        (server_cert_file.path(), &server_cert),
+        (server_key_file.path(), &server_key),
+        (client_ca_file.path(), &server_ca),
+        (client_cert_file.path(), &server_cert),
+        (client_key_file.path(), &server_key),
```
flavio, line 770: this is wrong, you should be generating a different client cert/key for mTLS instead of reusing the server one. They can be provided by a different root CA, hence it's fine to just do something like:

```rust
let (client_ca, client_cert, client_key) = generate_tls_certs();
```

flavio (changes requested):
I like it. The only comment that must be addressed is the one about the mTLS tests

viccuad (approved):
LGTM, modulo Flavio's suggestions.

src/config.rs
```diff
@@ -164,7 +162,6 @@ impl Config {
             log_level,
             log_fmt,
             log_no_color,
-            otlp_endpoint,
```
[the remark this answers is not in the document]
viccuad: Seems to be the case. Moreover, we set the env var in the test.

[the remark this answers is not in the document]
fabriziosestito: yes exactly

jvanz:
:+1:

PR #1016, opened 2025-01-03
fix!: renames "url" field into "module".

### LINKED ISSUE #906: policies.yaml should match CRDs: rename url field to module.

The PolicyServer config should match the format of policies defined in our CRDs.
For this reason, we might want to rename the `url` field in `module`. 
This also means changing the `ConfigMap` generation in the controller, by removing the configuration translation logic. 

NOTE: This is a breaking change, but it will impact only users using a PolicyServer without the controller, such as raw policy users. 

### Solution you'd like

_No response_

_No response_

### Anything else?

_No response_

jvanz:
## Description

Renames the "url" field into "module". This makes the field name the same of the name used in the CRDs.

Fix #906

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -318,7 +318,7 @@ pub enum PolicyOrPolicyGroupSettings {
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 pub struct PolicyGroupMember {
     /// Thge URL where the policy is located
-    pub url: String,
+    pub module: String,
     /// The settings for the policy
     pub settings: Option<HashMap<String, serde_yaml::Value>>,
     /// The list of Kubernetes resources the policy is allowed to access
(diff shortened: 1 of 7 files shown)


viccuad (changes requested):
Looks good! Yet we should change also `policies.yml.example`.

viccuad (approved):

src/config.rs
```diff
@@ -341,7 +341,7 @@ pub enum PolicyOrPolicyGroup {
     #[serde(rename_all = "camelCase")]
     Policy {
         /// The URL where the policy is located
-        url: String,
+        module: String,
```
fabriziosestito, line 344: what about using `#[serde(alias = "url")]` so we do not break the current behavior?

fabriziosestito: Thinking it over, this only affects raw policy users, so I'm fine with the breaking changes.

fabriziosestito (approved):

src/config.rs
viccuad: Opened https://github.com/kubewarden/docs/pull/484 for updating the user-facing docs.

Issue #810, opened 2024-07-04
Feature: automatically create docs for `policy-server` commands and flags


viccuad:
OpenSSF best practices asks us to document our API. As cli tool, that means `kwctl` commands and flags.

### Acceptance criteria

Add a new `kwctl docs`, that outputs a Markdown or similar file with the docs for all our clap commands and flags. This can be achieved with [clap-markdown](https://docs.rs/clap-markdown/latest/clap_markdown/).

On release, and after building kwctl, run `kwctl docs` and ship the resulting markdown file as GH artifact in the GH release.

Use updatecli to consume the new markdown doc file from the GH release, and automate its inclusion into the reference section of kubewarden/docs.

jvanz:
@kubewarden/kubewarden-developers decided to merge this in `v1.21` release.

PR #1056, opened 2025-01-27
feat(cli): generate CLI reference documentation.

jvanz:
## Description

Adds the "docs" subcommand to allow the CLI reference documentation generation in markdown format.

Related to #810

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,10 +1,12 @@
 mod cli;
 
 use std::fs;
+use std::io::prelude::*;
 
 use ::tracing::info;
 use anyhow::anyhow;
 use anyhow::Result;
+use clap::ArgMatches;
 use opentelemetry::global::shutdown_tracer_provider;
 use policy_server::metrics::setup_metrics;
 use policy_server::tracing::setup_tracing;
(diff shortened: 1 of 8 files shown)


src/main.rs
```diff
@@ -19,40 +20,55 @@ async fn main() -> Result<()> {
         .expect("Failed to install crypto provider");
 
     let matches = cli::build_cli().get_matches();
-    let config = policy_server::config::Config::from_args(&matches)?;
+    match matches.subcommand_name() {
```
flavio, line 23: I would prefer to avoid the whole indentation caused by this change.

I think doing something like that would leave the code more readable:

```
let matches = cli::build_cli().get_matches();
if matches.subcommand_name() == Some("docs") {
   // maybe use a different name for that function, that's the 1st thing that came to my mind
   return handle_docs_subcommand(&matches);
}

// leave the code unchanged

flavio (changes requested):
This looks good and I would even try to merge it before we tag 1.21.0 since it's a really small change.

I just left one minor code change request

viccuad (approved):
LGTM modulo Flavio's suggestion.

>  I would even try to merge it before we tag 1.21.0 since it's a really small change.

I agree.

fabriziosestito (approved):

jvanz:
As we are almost releasing the v1.21. I'm moving this to blocked to be merged after the v1.21 release

jvanz:
@kubewarden/kubewarden-developers decided to merge this in `v1.21`.

PR #1070, opened 2025-02-05
fix: address broken sigstore's TUF repository

flavio:
The contents of Sigstore's TUF repository have been recently changed and, due to a mistake, the sigstore-rs was broken.

This commit consumes a forked version of sigstore-rs, which consumes a forked version of the tough crate.

This is required to fix https://github.com/sigstore/sigstore-rs/issues/429 until upstream changes the contents of TUF's repository.


flavio:
The CI tests failed because of a glitch inside of the otel test. We can retrigger them, or just move ahead

viccuad (approved):

Issue #1067, opened 2025-02-04
Add mTLS support


fabriziosestito:
The policy-server has a new flag that enables mTLS. The CA certificate is read from the local filesytem and is then used to enforce mTLS.
The policy-server should also monitor this file for changes and automatically reload the certificate when it changes.

Admission criteria:
- [x] Introduce new CLI flag that reads the client CA file from disk. https://github.com/kubewarden/policy-server/pull/1075
- [x] #1077
- [x] #1078

jvanz:
All the policy server changes have been addressed. Closing this issue.

PR #1075, opened 2025-02-07
feat: enable mTLS.

jvanz:
## Description

Updates policy server adding a new CLI flag that can be used to enable mTLS. If the new CLI flag, or environment variable, setting a client CA file is defined, policy server will enable mTLS. Thus, it will accept request only from clients where the certificate is signed by the give CA.

Part of #1067


fabriziosestito (changes requested):

tests/integration_test.rs
```diff
@@ -677,17 +686,23 @@ async fn test_detect_certificate_rotation() {
     let certs_dir = tempfile::tempdir().unwrap();
     let cert_file = certs_dir.path().join("policy-server.pem");
     let key_file = certs_dir.path().join("policy-server-key.pem");
+    let client_cert_file = certs_dir.path().join("client_cert.pem");
```
fabriziosestito, line 689: I understand this is not a CA, but we are using it as a client CA in the mTLS configuration. I believe this is fine within the scope of this test. However, could we rename the variable and the file name to `client-ca(.pem)`?

tests/integration_test.rs
```diff
@@ -677,17 +686,23 @@ async fn test_detect_certificate_rotation() {
...
+    let client_key_file = certs_dir.path().join("client_key.pem");
```
fabriziosestito, line 690: we can ignore the key file, there is no need to write this.

tests/integration_test.rs
```diff
@@ -721,6 +743,7 @@ async fn test_detect_certificate_rotation() {
 
     let hostname2 = "cert2.example.com";
     let tls_data2 = create_cert(hostname2);
+    let tls_data_client2 = create_cert(hostname2);
```
fabriziosestito, line 746: `client_ca2` is clearer

src/cli.rs
```diff
@@ -88,6 +88,13 @@ pub(crate) fn build_cli() -> Command {
             .env("KUBEWARDEN_KEY_FILE")
             .help("Path to an X.509 private key file for HTTPS"),
 
+        Arg::new("client-ca-cert-file")
+            .long("client-ca-cert-file")
+            .value_name("CLIENT_CA_CERT_FILE")
+            .default_value("")
+            .env("KUBEWARDEN_CLIENT_CA_FILE")
+            .help("Path to an X.509 certificate file for mTLS"),
```
flavio, line 96: Also, this help message is wrong. This is the CA that issued the client certificat

flavio (commented):
There's still some work to be done, but I think it should be handled with 2 separate PRs. I'll create dedicated sub-issues

src/config.rs
```diff
@@ -182,14 +176,18 @@ fn api_bind_address(matches: &clap::ArgMatches) -> Result<SocketAddr> {
     .map_err(|e| anyhow!("error parsing arguments: {}", e))
 }
 
-fn tls_files(matches: &clap::ArgMatches) -> Result<(String, String)> {
+fn tls_files(matches: &clap::ArgMatches) -> Result<TlsConfig> {
```
flavio, line 179: rename that to something like `build_tls_config`

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+// service certificates. Therefore, we need to load everything and build the ServerConfig
+async fn build_tls_server_config(tls_config: &TlsConfig) -> Result<rustls::ServerConfig> {
+    use std::{fs::File, io::BufReader};
+
+    use rustls::{server::WebPkiClientVerifier, RootCertStore, ServerConfig};
+    use rustls_pemfile::Item;
+    use rustls_pki_types::{CertificateDer, PrivateKeyDer};
+
+    let cert_file = &mut BufReader::new(File::open(tls_config.cert_file.clone()).unwrap());
+    let key_file = &mut BufReader::new(File::open(tls_config.key_file.clone()).unwrap());
```
flavio, line 287: I don't like the `unwrap`. Please raise an error and state which file could not be opened too

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+    }
+
+    let cert = cert.into_iter().map(CertificateDer::from).collect();
+    let key = PrivateKeyDer::try_from(key_vec.pop().unwrap())
+        .map_err(|e| anyhow!("Cannot parse server key: {e}"))?;
+
+    let config = if let Some(client_ca_cert_file_path) = tls_config.client_ca_cert_file.clone() {
+        // we have the client CA. Therefore, we should enable mTLS.
+        let client_ca_cert_file =
+            &mut BufReader::new(File::open(client_ca_cert_file_path).unwrap());
```
flavio, line 310: I don't like the `unwrap`. Please raise an error and state which file could not be opened too

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+
+    use rustls::{server::WebPkiClientVerifier, RootCertStore, ServerConfig};
+    use rustls_pemfile::Item;
+    use rustls_pki_types::{CertificateDer, PrivateKeyDer};
+
+    let cert_file = &mut BufReader::new(File::open(tls_config.cert_file.clone()).unwrap());
+    let key_file = &mut BufReader::new(File::open(tls_config.key_file.clone()).unwrap());
+    let cert = rustls_pemfile::certs(cert_file)
+        .map(|it| it.map(|it| it.to_vec()))
+        .collect::<Result<Vec<_>, _>>()?;
```
flavio, line 290: a vector of vector looks ugly, try using `flatten`, also... the PEM file should contain only 1 certificate

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+    let mut key_vec: Vec<Vec<u8>> = rustls_pemfile::read_all(key_file)
+        .filter_map(|i| match i.ok()? {
+            Item::Sec1Key(key) => Some(key.secret_sec1_der().to_vec()),
+            Item::Pkcs1Key(key) => Some(key.secret_pkcs1_der().to_vec()),
+            Item::Pkcs8Key(key) => Some(key.secret_pkcs8_der().to_vec()),
+            _ => None,
```
flavio, line 296: I would log that, it's odd to put a something that is not a key into a pem file that is supposed to contain a key:

```rust
_ => {
   info!("found something that is not a key, ignoring");
   None
}
```

Obviously with a better message :)

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+        })
+        .collect();
+    if key_vec.len() != 1 {
+        return Err(anyhow!("Cannot parse server key"));
+    }
```
flavio, line 301: I  would do something more sophisticated:

```rust
if key_vec.len() == 0 {
   return Err(anyhow!("no key provided"));
}
if key_vec.len() > 1 {
   return Err(anyhow!("PEM file contained more than one key"));
}
```

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+        .map_err(|e| anyhow!("Cannot parse server key: {e}"))?;
+
+    let config = if let Some(client_ca_cert_file_path) = tls_config.client_ca_cert_file.clone() {
+        // we have the client CA. Therefore, we should enable mTLS.
+        let client_ca_cert_file =
+            &mut BufReader::new(File::open(client_ca_cert_file_path).unwrap());
+
+        let mut ca_certs = RootCertStore::empty();
+        let client_ca_certs: Vec<_> = rustls_pemfile::certs(client_ca_cert_file)
+            .filter_map(|it| it.ok())
```
flavio, line 314: Here we are silently skipping all the certificates that errored while being read.

I would do something like:

```rust
.filter_map(|it| {
  if let Err(e) = it {
     warn!(error=e, "cannot convert client CA file, ignoring")
  }
  it.ok()
})
```

src/lib.rs
```diff
@@ -271,6 +271,62 @@ impl PolicyServer {
...
+            .collect();
+        ca_certs.add_parsable_certificates(client_ca_certs);
```
flavio, line 316: Maybe add:

```rust
info!(client_ca_certs=client_ca_certs.size(), "adding client CA certificates for mTLS");
```

src/lib.rs
```diff
@@ -293,24 +349,36 @@ async fn create_tls_config_and_watch_certificate_changes(
 ) -> Result<RustlsConfig> {
     use ::tracing::error;
 
-    let cert_file = tls_config.cert_file.clone();
-    let key_file = tls_config.key_file.clone();
+    let cert_file_path = tls_config.cert_file.clone();
+    let key_file_path = tls_config.key_file.clone();
+    let client_ca_cert_file = tls_config.client_ca_cert_file.clone();
```
flavio, line 354: nitpicking: call it like the other siblings: `client_ca_cert_path`

flavio (changes requested):

fabriziosestito (approved):
LGTM, we agreed on adding a test, tracked in  #1085

Issue #1077, opened 2025-02-10
Expose readiness probe over http


flavio:
With the introduction of mTLS, the readiness endpoint would become unreachable to Kubernetes.

As a solution we have to:

- Create two different axum routers: the existing one minus the `/readiness` endpoint, and a new one with only the `/readiness` probe
- Start two web servers (one per router):
  - Readiness web server: listens always on plain http
  - Validation web server: listens on http or on https depending on how the user started the policy-server. mTLS, if enabled, is enforced only on this server

Issue #1078, opened 2025-02-10
mTLS: handle multiple CAs


flavio:
There are two clients interacting with the `/validate` endpoints of Policy Server: the kubernetes API server and `audit-scanner`. When mTLS is enabled they are going to use client certificates issued by two different CA.

The following changes have to be done:
- Allow the `--client-ca-file` to take multiple values by using a comma separator. This must be handled also when reading the contents of the environment variable
- Extend the inotify code to watch for changes done to all the files mentioned by the `--client-ca-file`

Talking about the inotify watch code, the code provided as part of https://github.com/kubewarden/policy-server/pull/1075/ has to be refactored a bit. The current code is not going to scale nicely (in terms of readability) when multiple client CAs are going to be handled.

PR #1080, opened 2025-02-11
feat: expose readiness probe over http

fabriziosestito:
## Description
Fix: #1077

Splits the api and the readiness server in two axum servers running on different ports.
[tokio::Notify](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) was used to start serving the readiness probe server right after the main API server.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,7 @@ use profiling::activate_memory_profiling;
 use rayon::prelude::*;
 use std::{fs, net::SocketAddr, sync::Arc};
 use tokio::{
-    sync::{oneshot, Semaphore},
+    sync::{oneshot, Notify, Semaphore},
     time,
 };
 use tower_http::trace::{self, TraceLayer};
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -225,40 +224,56 @@ impl PolicyServer {
...
+                notify.notify_one();
+
+                server.serve(self.router.into_make_service()).await
+            }
+        };
+
+        let readiness_probe_server = async {
+            notify.notified().await;
+
+            axum_server::bind(self.readiness_probe_addr)
```
flavio, line 267: I would add some logging telling the user the `/readiness` probe is ready and listening over port XYZ

flavio (approved):

jvanz (approved):

viccuad (approved):

jvanz:
@fabriziosestito I've rebased this PR to allow me to merge it. I want to write some tests for the mTLS after my changes in #1078.

PR #1087, opened 2025-02-12
fix: TlsConfig and mTLS logic

fabriziosestito:
## Description  

With the latest changes, `TlsConfig` was always set to `Some`, causing the policy server to attempt loading certificate files from the default CLI value, which was an empty string (`""`). As a result, the policy server failed to start.  

This PR introduces an additional validation step:  
1. If either the certificate or key is provided while the other is missing.
2. If neither a certificate nor a key is provided via flags, `TlsConfig` is set to `None`.  
3. If a client CA is specified but the certificate and key are missing, an error is raised.  

The rationale behind point 3 is that mTLS is only enabled when server certificates are configured.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -130,7 +130,7 @@ impl Config {
             .expect("clap should have assigned a default value")
             .to_owned();
 
-        let tls_config = Some(build_tls_config(matches)?);
+        let tls_config = build_tls_config(matches)?;
 
         let enable_pprof = matches
             .get_one::<bool>("enable-pprof")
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -189,18 +189,28 @@ fn readiness_probe_bind_address(matches: &clap::ArgMatches) -> Result<SocketAddr
...
+        // No TLS configuration provided
+        (None, None, None) => Ok(None),
+        // Client CA certificate provided without server certificate and key
+        (None, None, Some(_)) => Err(anyhow!(
+            "client CA certificate requires server certificate and key to be specified"
+        )),
+        // Server certificate or key provided without the other
+        (Some(_), None, _) | (None, Some(_), _) => Err(anyhow!(
+            "both certificate and key must be provided together"
+        )),
```
flavio, line 212: I like this "elixir-style" pattern matching!

flavio (approved):

jvanz (approved):

Issue #1091, opened 2025-02-14
Fix certs hot-reload logic and cleanup


fabriziosestito:
## Current Behavior

When client CA certificates are updated, the entire `ServerConfig` is reloaded. This can lead to potential race conditions if this reload happens concurrently with server certificate or key updates, potentially resulting in temporary invalid certificate states.

## Proposed Changes

Keep existing server certificates while reloading client CAs to prevent inconsistent states

Refactor the certificate management code:

- Split `build_tls_server_config` into smaller functions 
- Replace blocking `std::fs` operations with async `tokio::fs`
- (optional) Introduce a new CertReloader type to encapsulate reload logic and state management

PR #1094, opened 2025-02-17
fix: remove TLS hot-reload race condition

fabriziosestito:
Fix: #1091

This PR resolves a race condition that occurs during TLS certificate reloads by retaining the previous certificate and key pair if the client CA file changes. This approach enables reloading the server certificate and client CA independently.

Additionally, it replaces blocking std::fs functions within async blocks with their asynchronous counterparts from tokio::fs.

### DIFF
--- a/src/certs.rs
+++ b/src/certs.rs
@@ -1,0 +1,219 @@
+use ::tracing::{info, warn};
+use anyhow::{anyhow, Result};
+use rustls::{server::WebPkiClientVerifier, RootCertStore, ServerConfig};
+use rustls_pemfile::Item;
+use rustls_pki_types::{CertificateDer, PrivateKeyDer};
+use std::{fs::File, io::BufReader, sync::Arc};
+
+// This is required by certificate hot reload when using inotify, which is available only on linux
+#[cfg(target_os = "linux")]
+use tokio_stream::StreamExt;
+
+use crate::config::TlsConfig;
+
+async fn load_server_cert_and_key(
+    cert_file: String,
+    key_file: String,
+) -> Result<(Vec<CertificateDer<'static>>, PrivateKeyDer<'static>)> {
+    let cert_reader = &mut BufReader::new(File::open(cert_file)?);
+    let cert: Vec<CertificateDer> = rustls_pemfile::certs(cert_reader)
+        .filter_map(|it| {
+            if let Err(ref e) = it {
+                warn!("Cannot parse certificate: {e}");
+                return None;
+            }
+            it.ok()
+        })
+        .collect();
+    if cert.len() > 1 {
+        return Err(anyhow!("Multiple certificates provided in cert file"));
+    }
+
+    let key_file_reader = &mut BufReader::new(File::open(key_file)?);
+    let mut key_vec: Vec<Vec<u8>> = rustls_pemfile::read_all(key_file_reader)
+        .filter_map(|i| match i.ok()? {
+            Item::Sec1Key(key) => Some(key.secret_sec1_der().to_vec()),
+            Item::Pkcs1Key(key) => Some(key.secret_pkcs1_der().to_vec()),
+            Item::Pkcs8Key(key) => Some(key.secret_pkcs8_der().to_vec()),
+            _ => {
+                info!("Ignoring non-key item in key file");
+                None
+            }
+        })
+        .collect();
+    if key_vec.is_empty() {
+        return Err(anyhow!("No key provided in key file"));
+    }
+    if key_vec.len() > 1 {
+        return Err(anyhow!("Multiple keys provided in key file"));
+    }
+    let key = PrivateKeyDer::try_from(key_vec.pop().unwrap())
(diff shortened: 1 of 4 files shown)


src/certs.rs
```diff
@@ -182,7 +180,12 @@ async fn load_server_cert_and_key(
     cert_file: &Path,
     key_file: &Path,
 ) -> Result<(Vec<CertificateDer<'static>>, PrivateKeyDer<'static>)> {
-    let cert_reader = &mut BufReader::new(File::open(cert_file)?);
+    let cert_contents = tokio::fs::read(cert_file).await?;
+    let key_contents = tokio::fs::read(key_file).await?;
+
+    let cert_reader = &mut BufReader::new(&cert_contents[..]);
+    let key_reader = &mut BufReader::new(&key_contents[..]);
```
fabriziosestito, line 187: since `rustls_pemfile`expectes a `std BufReader` we need to load the contents in memory and then create one to avoid blocking the tokio runtime with `std::fs` calls.

tests/integration_test.rs
```diff
@@ -728,7 +732,7 @@ async fn test_detect_certificate_rotation() {
     let tls_data2 = create_cert(hostname2);
 
     // write only the cert file
-    std::fs::write(&cert_file, tls_data2.cert.clone()).unwrap();
+    fs::write(&cert_file, tls_data2.cert.clone()).await.unwrap();
```
fabriziosestito, line 735: let's always use `tokio::fs` in async blocks

src/certs.rs
```diff
@@ -187,32 +105,146 @@ pub(crate) async fn create_tls_config_and_watch_certificate_changes(
...
             }
 
-            // if both the certificate and the key have been changed or there is no change in the
-            // server cert and key, but the client cert changed, reload the certificate
-            if (key_changed && cert_changed)
-                || (client_cert_changed && (key_changed == cert_changed))
-            {
-                info!("Reloading TLS certificates");
+            // Reload the client CA certificates if they have changed, keeping the current server certificates unchanged
+            if client_ca_changed {
```
fabriziosestito, line 116: If an inotify event indicates a change in the client CA, only the client verifier should be reloaded. If the TLS server certificate or key were modified simultaneously, but the inotify event for those changes had not yet arrived, we could end up with a broken TLS configuration till the next inotify event. Reloading only the client verifier in this scenario avoids partial updates of the certificate and key.

src/certs.rs
```diff
@@ -0,0 +1,258 @@
...
+        .into_iter()
+        .map(|path| {
+            inotify
+                .watches()
+                .add(path, inotify::WatchMask::CLOSE_WRITE)
+                .map_err(|e| anyhow!("Cannot watch client certificate file: {e}"))
+        })
+        .collect();
+
+    let client_ca_watches = client_ca_watches?;
```
jvanz, line 79: Why not merge these two lines into one?

src/certs.rs
```diff
@@ -0,0 +1,258 @@
...
+
+            // Reload the client CA certificates if they have changed, keeping the current server certificates unchanged
+            if client_ca_changed {
+                info!("Reloading Client CA certificates");
+
+                client_ca_changed = false;
+
+                match load_client_ca_certs(tls_config.client_ca_file.clone()).await {
+                    Ok(cv) => {
+                        client_verifier = Some(cv);
```
jvanz, line 125: I think you are using a variable from outside the scope of the task because the variable is moved to the task scope. Right? Have you did that intentionally, right? The same apply for the `cert` and `key` variable some lines below. Just double checking. No problem for me so far.

src/certs.rs
jvanz: I just saw your commit removing unwraps. :+1:

jvanz:
I've just rebased this PR on top of main after merging base PR for these changes #1090

src/certs.rs
fabriziosestito: The variable is the moved outer scope `mut` variable. The outer scope performs the first initialization. 
This is needed to preserve the state of the certificates/client CA so we can perform a partial reload.

src/certs.rs
```diff
@@ -0,0 +1,258 @@
...
+async fn create_tls_config_and_watch_certificate_changes(
+    tls_config: TlsConfig,
+) -> Result<RustlsConfig> {
+    let cfg = RustlsConfig::from_pem_file(tls_config.cert_file, tls_config.key_file).await?;
+    Ok(cfg)
+}
+
+/// Return the RustlsConfig and watch for changes in the certificate files
+/// using inotify.
+/// When a both the certificate and its key are changed, the RustlsConfig is reloaded,
```
viccuad, line 26:
```suggestion
/// When both the certificate and its key are changed, the RustlsConfig is reloaded,
```

viccuad (approved):

src/certs.rs
jvanz: Yes, I'm fine with that. As I said, I was just double checking if that was on purpose.

jvanz (approved):

PR #1145, opened 2025-03-24
feat: Skip Wasm Evaluation for Requests from 'Always Allowed' Namespaces and Improve Metrics Handling

### LINKED ISSUE #599: Requests from "always allowed" namespace should not be sent to Wasm evaluators
Policy Server administrator can configure a special namespace that is not affected by the policies, meaning all the operations inside of it are always allowed. This is commonly used to exclude Kubewarden's namespace, preventing user's policies from damaging the whole stack (and hence causing an outage to the whole cluster).

Currently, the requests assigned to this namespace are still processed by the WebAssembly evaluators, but later the evaluation results are simply discarded and the request is accepted. As a minor optimization, we should change our code to immediately accept a request involving the "special namespace", without having to waste CPU cycles doing WebAssembly evaluations.

## Admission Criteria

- Change the policy server code to immediately accept requests involving one of the namespaces that is on the "always allow" list.

maanugh:
…ces and Improve Metrics Handling

## Description

Fix #599 

## Test

### DIFF
--- a/src/api/service.rs
+++ b/src/api/service.rs
@@ -36,6 +36,40 @@ pub(crate) fn evaluate(
     let start_time = Instant::now();
     let policy_id: PolicyID = policy_id.parse()?;
 
+    // Early check for requests from special namespaces
+    if let ValidateRequest::AdmissionRequest(adm_req) = validate_request {
+        if let Some(ref req_namespace) = adm_req.namespace {
+            if evaluation_environment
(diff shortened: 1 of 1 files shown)


flavio (approved):
Thanks for your contribution, that looks good to me.

Let's see what the other maintainers think about it

viccuad (approved):
LGTM, thanks for the contribution.

fabriziosestito (approved):
LGTM

PR #1150, opened 2025-03-26
fix: removed the rustls=off filter directive, which was previously suppressing all TLS-related message

### LINKED ISSUE #1148: Feature Request:

policy-server silently fails requests missing valid TLS certificates.

This is slightly different than the kubewarden-controller pod, which logs such HTTPS out of the box:

```console
2025/03/24 16:37:06 http: TLS handshake error from 10.42.0.10:55828: tls: client didn't provide a certificate                                                                             
```

### Solution you'd like

Ideally both would log such requests, so that users can more easily detect abuse and/or misconfigured clients.

_No response_

### Anything else?

Maybe we're being too aggressive filtering the log messages. See [these lines](https://github.com/kubewarden/policy-server/blob/9ccdbc246e7bbdf17deb3f8620f921ae15ce5239/src/tracing.rs#L27-L29).

maanugh:
## Description

Fix #1148 

## Test

### DIFF
--- a/src/tracing.rs
+++ b/src/tracing.rs
@@ -26,7 +26,6 @@ pub fn setup_tracing(
         .add_directive("cranelift_wasm=off".parse().unwrap())
         .add_directive("h2=off".parse().unwrap())
         .add_directive("hyper=off".parse().unwrap())
-        .add_directive("rustls=off".parse().unwrap())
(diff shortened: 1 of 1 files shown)


viccuad (approved):
Good find, thanks for this fix!

maanugh:
Failed connections will appear at log level INFO.

PR #1166, opened 2025-04-25
feat: check ```protocol_version``` before storing

### LINKED ISSUE #642: Check for policies' `protocol_version`  and error if protocol version is not supported.
### Acceptance criteria

Policy server should consume `policy_evaluator::evaluator::PolicyEvaluator::protocol_version()`, and error if the policy protocol version is not supported.

See kwctl for an example: https://github.com/kubewarden/kwctl/blob/23c4eae60d5a687bc64382b2cae3cd725e529637/src/backend.rs#L52-L61

manukirat:
## Description

Fix #642 

## Test

### DIFF
--- a/src/evaluation/precompiled_policy.rs
+++ b/src/evaluation/precompiled_policy.rs
@@ -1,7 +1,7 @@
 use anyhow::{anyhow, Result};
 use lazy_static::lazy_static;
 use policy_evaluator::{
-    policy_evaluator::PolicyExecutionMode, policy_metadata::Metadata, wasmtime,
+    policy_evaluator::PolicyExecutionMode, policy_metadata::Metadata, wasmtime, ProtocolVersion,
 };
 use semver::{BuildMetadata, Prerelease, Version};
 use sha2::{Digest, Sha256};
@@ -50,6 +50,8 @@ impl PrecompiledPolicy {
         let execution_mode = metadata.execution_mode;
         has_minimum_kubewarden_version(&metadata)?;
 
+        has_valid_protocol_version(&metadata)?;
+
         let precompiled_module = engine.precompile_module(&policy_contents)?;
 
         let mut hasher = Sha256::new();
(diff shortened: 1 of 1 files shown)


viccuad:
Moving to blocked so we don't merge meanwhile we are on RC phase.

flavio:
I looked at the PR and I started to wonder what's the purpose of doing this check. I mean, invoking `protocol_version()` against a policy and make sure it doesn't return an `Error` is fine, but what are we supposed to do with the actual contents of the `Result`?

I think we have to change approach entirely.

Inside of [PrecompiledPolicy::new](https://github.com/kubewarden/policy-server/blob/34cda5a514d2ae03c7113d5d4cb6f471798f748d/src/evaluation/precompiled_policy.rs#L51), we have to add check.
The check must be done only if the execution mode of the policy is wapc (that's the only mode that implements the protocol_version). We have to ensure we can obtain a protocol version (there's no error) and that the protocol version is equal to `V1` (there's an [enum for that](https://docs.rs/kubewarden-policy-sdk/0.13.3/kubewarden_policy_sdk/metadata/enum.ProtocolVersion.html))

flavio (changes requested):
I looked at the PR and I started to wonder what's the purpose of doing this check. I mean, invoking `protocol_version()` against a policy and make sure it doesn't return an `Error` is fine, but what are we supposed to do with the actual contents of the `Result`?

I think we have to change approach entirely.

Inside of [PrecompiledPolicy::new](https://github.com/kubewarden/policy-server/blob/34cda5a514d2ae03c7113d5d4cb6f471798f748d/src/evaluation/precompiled_policy.rs#L51), we have to add check.
The check must be done only if the execution mode of the policy is wapc (that's the only mode that implements the protocol_version). We have to ensure we can obtain a protocol version (there's no error) and that the protocol version is equal to `V1` (there's an [enum for that](https://docs.rs/kubewarden-policy-sdk/0.13.3/kubewarden_policy_sdk/metadata/enum.ProtocolVersion.html))

maanugh:
ok, got it.

flavio (changes requested):
LGTM, but please write some unit tests. Bonus points if you use rstest to handle the different cases

maanugh:
added the test cases.

src/evaluation/precompiled_policy.rs
```diff
@@ -141,4 +156,24 @@ mod tests {
...
+
+    #[rstest]
+    #[case(Metadata {
+        execution_mode: PolicyExecutionMode::KubewardenWapc,
+        protocol_version: Some(ProtocolVersion::Unknown),
+        ..Default::default()
+    })]
+    fn invalid_protocol_version_test(#[case] metadata: Metadata) {
+        assert!(has_valid_protocol_version(&metadata).is_err())
+    }
```
flavio, line 178: I meant to write them like that:

```rust
    #[rstest]
    #[case(Metadata {
        execution_mode: PolicyExecutionMode::KubewardenWapc,
        protocol_version: Some(ProtocolVersion::Unknown),
        ..Default::default()
    }, false)]
    fn validate_protocol_version_test(#[case] metadata: Metadata, #[case] is_valid: bool) {
        assert_eq!(is_valid, has_valid_protocol_version(&metadata).is_err())
    }
```

Also, write a test case for non-wapc policies.

jvanz (approved):

PR #1188, opened 2025-06-03
fix: Only check for ProtocolVersion on Wapc execution mode

viccuad:
## Description

Fix https://github.com/kubewarden/helm-charts/issues/689

## Test

CI. Amended relevant test.


flavio (approved):
Thanks for the fix, I feel sorry for not having spotted this regression :(

PR #1194, opened 2025-06-11
feat: custom error message support.

### LINKED ISSUE #1190: Add support for custom error message in policy evaluations
- [ ] Update the `policies.yml` Rust structure to take into account the new message
  - [ ] Override the policy failure message when use provided his own custom message
  - [ ] [OPTIONAL for 1st iteration]: add the original rejection message as a warning. Do the same thing done by group policies to track policy member rejections
- [ ] File an issue to the Rancher UI Extension: the UI has to take into account the new `spec.message` field that is now available to `ClusterAdmissionPolicy` and `AdmissionPolicy`

jvanz:
## Description

Updates the policy server to allow user define a "message" field in the policies.yml file to customize the error message returned by the policy.

Fix #1190

### DIFF
--- a/src/evaluation/policy_evaluation_settings.rs
+++ b/src/evaluation/policy_evaluation_settings.rs
@@ -11,4 +11,6 @@ pub(crate) struct PolicyEvaluationSettings {
     pub(crate) allowed_to_mutate: bool,
     /// The policy-specific settings provided by the user
     pub(crate) settings: PolicyOrPolicyGroupSettings,
+    /// Determines a custom rejection message for the policy
+    pub(crate) custom_rejection_message: Option<String>,
 }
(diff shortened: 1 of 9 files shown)


src/evaluation/policy_evaluation_settings.rs
```diff
@@ -11,4 +11,6 @@ pub(crate) struct PolicyEvaluationSettings {
     pub(crate) allowed_to_mutate: bool,
     /// The policy-specific settings provided by the user
     pub(crate) settings: PolicyOrPolicyGroupSettings,
+    /// Determines a custom error bessage for the policy
+    pub(crate) custom_error_message: Option<String>,
```
jvanz, line 15: Should change the name of this field to message? Or change the `message` field in the policies.yml into `customer_error_message`. It like the later because it makes more clear the goal of the configuration.

PR #1211, opened 2025-07-11
chore: update policy-evaluator.

jvanz:
## Description

Updates the policy-evaluator dependency to v0.27.1


viccuad (approved):
Weird that the workflows need approval.

Issue #1270, opened 2025-09-03
flaky tests of timeoutSeconds and timeoutEvalSeconds


viccuad:
After the addition [here](https://github.com/kubewarden/policy-server/pull/1250) of a new integration test for timeoutEvalSeconds,
the [3 timeout tests are flaky](https://github.com/kubewarden/policy-server/blob/main/tests/integration_test.rs#L421-L515). The failure shifts between on of the 3.

There was a preliminary investigation done, see comments starting [here](https://github.com/kubewarden/policy-server/pull/1250#issuecomment-3233078525).

This flakiness manifests in all CI jobs, needing several runs to pass (annoying on releases). Example [here](https://github.com/kubewarden/policy-server/actions/runs/17421418535?pr=1268) (4 runs and counting).

### Acceptance criteria

Fix the flaky tests, preferably without increasing the sleep timer of the sleep policy, since it is honored fully (see [docs](https://docs.kubewarden.io/reference/policy-evaluation-timeout#limitations)) and will increase the duration of the testsuite.

PR #1284, opened 2025-09-19
fix: Correctly create PolicyEvaluatorPre and consume timeouts for spec.timeoutEvalSeconds feature

viccuad:
## Description

Fix https://github.com/kubewarden/policy-server/issues/1270.
With this, both integration tests and coverage (which run integration tests) should not be flaky anymore.

This PR fixes 2 bugs:

1. Ensure we enable eval timeout even if we don't have policy-server global timeout.

2. Before, we were creating a PolicyEvaluator per compiled module.
This is not enough, as each policy can now have a different
spec.timeoutEvalSeconds.
This bug made the timeout integration test fail in a flaky manner.
The integration tests have 2 policies, `sleep` and `sleep-1s-timeout`.
They are saved in a hashmap (nondeterministic order). If on the run the
`sleep` policy was registered first, the `sleep-1s-timeout` would reuse
the `sleep` PolicyEvaluator, which meant that it didn't have the
spec.timeoutEvalSeconds assigned and the test would fail.
To fix this:
    - Extend the check when registering and creating a PolicyEvaluatorPre
  from only the module to the module + policy_evaluation_limit_seconds.
  Add new `PolicyEvaluatorPreKey{}` with those values as key for the
  hashmap of policies and PolicyEvaluatorPre.
    - Add `module_digest_and_eval_timeout_to_policy_evaluator_pre`, needed
  to obtain the eval timeout used in the key when we are rehydrating.

## Test

Extended existing unit test for creation of PolicyEvaluators.
Use existing integration tests for the timeoutEvalSeconds feature.
Add configurable logging output to integration tests.

### Tradeoff

Irrelevant space cost increase of `+N` on worst case, N: number of policies. As we have a new hashmap of `policy_id`, `eval timeout`.


viccuad:
Dropped the commit adding logging to the integration tests, which slowed them in CI.

flavio:
I just did a force push because I had to solve some conflicts between this branch and `main`.

jvanz (approved):

PR #1303, opened 2025-10-17
fix sigstore handle multiple rekor keys

flavio:
Allow multiple Rekor keys to be found, and used, inside of Sigstore's TUF repository.


flavio:
This is also preparing for the 1.29.2 patch release

viccuad (approved):
LGTM

PR #1306, opened 2025-10-21
prepare for release

flavio:
Prepare the 1.30.0-rc1 release:

- **chore: update to rust edition 2024**: since we're already using rust 1.90.0, it makes a lot of sense to move to latest Rust edition. In this way we can use some syntax changes that make the code more readable. This change caused the usual amount of fmt and lint fixes
- **chore(deps): update to latest version of policy-evaluator**: needed to consume latest version of wasmtime
- **chore: prepare 1.30.0-rc1 release**: just bumping the version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4017,7 +4017,7 @@ dependencies = [
 
 [[package]]
 name = "policy-server"
-version = "1.29.2"
+version = "1.30.0-rc1"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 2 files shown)


jvanz (approved):

jvanz (changes requested):
Please, take a look in the failing CI. ;)

viccuad (approved):

PR #1320, opened 2025-11-14
update deps

flavio:
This PR updates wasmtime to latest patch release to address RUSTSEC-2025-0118.

It also updates the `cargo audit` ignore list, to remove all the issues that have been addressed.


viccuad (approved):
:pray: for updating the ignore list.

